PR #22, opened 2023-08-30
Deleting previous trained procedures

joker2411:
Changes to be committed:
	modified:   train.py
	modified:   utils.py

## Description of the change

> Deleted all the previous procedures whose names were starting with trained procedure name.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -84,9 +84,12 @@ def delete_import_files(session, stage_name, import_paths: List[str]) -> None:
 def delete_procedures(session, train_procedure) -> None:
     procedures = session.sql(f"show procedures like '{train_procedure}%'").collect()
     for row in procedures:
-        words = row.arguments.split(' ')[:-2]
-        procedure_arguments = ' '.join(words)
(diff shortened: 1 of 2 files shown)


utils.py
```diff
@@ -81,6 +81,13 @@ def delete_import_files(session, stage_name, import_paths: List[str]) -> None:
         if any(substring in row.name for substring in import_files):
             session.sql(f"remove @{row.name}").collect()
 
+def delete_procedures(session, train_procedure) -> None:
+    procedures = session.sql(f"show procedures like '{train_procedure}%'").collect()
+    for row in procedures:
+        words = row.arguments.split(' ')[:-2]
```
dpatchigolla, line 87: Can you keep this whole function in a try except block, with a `pass` in except? Splitting the arguments and taking till -2 args - that looks a bit fragile. As this is to tackle a corner case, it's better to let the function pass in case of an exception than to fail. The error message anyway is clear enough to point us in the right direction.

dpatchigolla (commented):
Added one comment

dpatchigolla (approved):

PR #25, opened 2023-09-01
Creating fn. for user specific feature-importance

joker2411:
Changes to be committed:
	modified:   config/data_prep.yaml
	modified:   predict.py
	modified:   train.py
	modified:   utils.py

## Description of the change

> created a function user-specific feature-importance table for top 5 and bottom 5 users as per their predictions.
> moved common function to utils script.


utils.py
```diff
@@ -480,3 +505,88 @@ def fetch_staged_file(session: snowflake.snowpark.Session,
...
+    explainer = shap.TreeExplainer(pipe['model'], feature_names=col_names_)
+    shap_score = explainer(pipe['preprocessor'].transform(data_x).astype(np.int_))
+    shap_df = pd.DataFrame(shap_score.values, columns=col_names_)
+    shap_df.insert(0, entity_column.upper(), data_y[entity_column.upper()].values)
+    shap_df.insert(len(shap_df.columns), score_column_name.upper(), data_y[score_column_name.upper()].values)
+    session.write_pandas(shap_df, table_name=f"Material_shopify_feature_importance", auto_create_table=True, overwrite=True)
+
+
+
+# ####  Functions for plotting feature importance score for single user. Might come in handy in future. ####
```
dpatchigolla, line 583: No need of commenting the function itself. Just add a comment saying it is not being called currently.

utils.py
```diff
@@ -480,3 +505,88 @@ def fetch_staged_file(session: snowflake.snowpark.Session,
...
+    fetch_staged_file(session, stage_name, file_name, target_folder)
+    if filetype == 'json':
+        f = open(os.path.join(target_folder, file_name), "r")
+        output_file = json.load(f)
+    elif filetype == 'joblib':
+        output_file = joblib.load(os.path.join(target_folder, file_name))
+    os.remove(os.path.join(target_folder, file_name))
+    return output_file
+
+def explain_prediction(creds, user_main_id, predictions_table_name, feature_table_name, predict_config):
```
dpatchigolla, line 519: This too is currently not being called anywhere right? Can you add that as a comment above the function?

dpatchigolla (commented):
inline comments.

utils.py
```diff
@@ -480,3 +505,88 @@ def fetch_staged_file(session: snowflake.snowpark.Session,
         with open(output_file_path, 'wb') as target_file:
             shutil.copyfileobj(gz_file, target_file)
     os.remove(input_file_path)
+
+def load_stage_file_from_local(session, stage_name, file_name, target_folder, filetype):
```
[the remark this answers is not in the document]
joker2411: 1st line in this function was already in use to save output image on local (inside _reports folder). So I have used that to fetch the files from stage and save it to local and created this new function to load those local files and delete those from local as well. This function can be used in future as well if we want to use this for other file types, we'll just keep on adding their functionalities in this function.

dpatchigolla (approved):

PR #29, opened 2023-09-06
Different users in train and test

joker2411:
Changes to be committed:
	modified:   train.py
	modified:   utils.py

## Description of the change

> saved sorted feature table on basis of ascending entity_column and descending index_ts.
> used that to filter out latest instances of all the unique users.
> corrected the latest_model_hash function.

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -1,4 +1,5 @@
 from sklearn.metrics import precision_recall_fscore_support, roc_auc_score, f1_score, average_precision_score,average_precision_score, PrecisionRecallDisplay, RocCurveDisplay, auc, roc_curve, precision_recall_curve
+from sklearn.model_selection import train_test_split
 import numpy as np 
 import pandas as pd
 from typing import Tuple, List, Union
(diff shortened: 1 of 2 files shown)


utils.py
```diff
@@ -300,17 +302,20 @@ def split_train_test(feature_table: snowflake.snowpark.Table,
...
-    train_x = X_train.drop(label_column, entity_column).to_pandas() # drop labels for training set
-    train_y = X_train.select(label_column).to_pandas()
-    test_x = X_test.drop(label_column, entity_column).to_pandas() # drop labels for training set
-    test_y = X_test.select(label_column).to_pandas()
-    val_x = X_val.drop(label_column, entity_column).to_pandas()
-    val_y = X_val.select(label_column).to_pandas()
+    session.write_pandas(X_train, table_name=f"{model_name_prefix}_train", auto_create_table=True, overwrite=True)
+    session.write_pandas(X_val, table_name=f"{model_name_prefix}_val", auto_create_table=True, overwrite=True)
+    session.write_pandas(X_test, table_name=f"{model_name_prefix}_test", auto_create_table=True, overwrite=True)
+    train_x = X_train.drop([entity_column.upper(), label_column.upper()], axis=1)
```
dpatchigolla, line 313: > entity_column.upper()

The column names in pandas get stored in upper case only if the snowflake table was created with uppercase (which is default). If they were created specifically as lowercase, the pandas df columns too would be lowercase and this will fail. I see that you used this a few more instances in the utils file. Please fix those too as a separate pr. For now, you can fix this alone. Handling them in snowflake tables is best way. 
If you want to make this work in pandas, write a separate function to correct the case sensitivity of inputs where you pass them through the snowflake table column names and return modified inputs.

dpatchigolla (commented):
Need to fix the case sensitivy.

PR #171, opened 2024-02-07
fix: update imports

shekhar-rudder:
## Description of the change

This PR updates the imports to run preprocess_and_train/predict files in remote mode

https://linear.app/rudderstack/issue/PRML-388/fix-imports-in-the-new-classifier-structure

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/ml_core/preprocess_and_train.py
```diff
@@ -301,6 +299,8 @@ def preprocess_and_train(
     if args.mode == constants.K8S_MODE:
         wh_creds_str = os.environ[constants.K8S_WH_CREDS_KEY]
         wh_creds = json.loads(wh_creds_str)
+    elif args.mode == constants.CI_MODE:
```
shekhar-rudder, line 302: Open to better ways of doing this

.github/workflows/unit-test.yaml
```diff
@@ -28,5 +28,11 @@ jobs:
     - name: Install dependencies
       run: pip install -r tests/integration/requirements.txt
 
+    - name: Compile core train
+      run: python -m src.ml_core.preprocess_and_train --mode ci
```
shekhar-rudder, line 32: Regression test

src/ml_core/preprocess_and_train.py
joker2411: This is better for current case I think as alternatively we can create a processor classes and objects here for each mode but processor classes needs session object for initiation which further needs _wh_creds_ coming from these modes. So, this creates kind of a circular dependency issue.

Right now at best, we can create a function in some other file and return the _wh_creds_ from that. Otherwise this is fine as well.

joker2411 (approved):

PR #176, opened 2024-02-09
Feature/prml 213 improvements to get material names

joker2411:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/utils/utils.py
```diff
@@ -469,9 +469,9 @@ def get_date_range(creation_ts: datetime, prediction_horizon_days: int) -> Tuple
     return str(start_date), str(end_date)
 
 
-def date_add(reference_date: str, add_days: int) -> str:
+def date_add(reference_date: str, add_days: int, **kwargs) -> str:
```
shekhar-rudder, line 472: Instead of adding a new flag `subtract` we can pass in a negative number in `add_days` argument

src/connectors/SnowflakeConnector.py
```diff
@@ -538,48 +557,106 @@ def get_material_names_(
...
                 label_snowpark_df.end_ts.alias("label_end_ts"),
             )
             for row in feature_label_snowpark_df.collect():
+                if (
+                    row.FEATURE_SEQ_NO
+                    and row.FEATURE_END_TS
+                    and row.LABEL_SEQ_NO
+                    and row.LABEL_END_TS
+                ):
+                    feature_seq_no = str(int(row.FEATURE_SEQ_NO))
```
shekhar-rudder, line 574: What error are you seeing if you use `feature_seq_no = row.FEATURE_SEQ_NO` ?

src/connectors/Connector.py
```diff
@@ -139,27 +158,37 @@ def generate_training_materials(
...
+            start_date,
+            features_profiles_model,
+            model_hash,
+            prediction_horizon_days,
         )
-        if materialise_feature_data and materialise_label_data:
+        materialise_feature_data = None
+        materialise_label_data = None
+        if feature_date is not None:
+            materialise_feature_data = utils.materialise_past_data(
```
shekhar-rudder, line 172: Is it possible for both `feature_date` and `label_date` to be not None ? If yes, we can create an array of 2 elements - feature and label date. Then we range over the array elements to call `materialise_past_data`. This way, we can get rid of one of the function calls

src/connectors/Connector.py
```diff
@@ -413,6 +446,18 @@ def get_default_label_value(
     ):
         pass
 
+    @abstractmethod
+    def get_valid_feature_label_dates(
+        self,
+        session,
+        feature_label_snowpark_df,
```
shekhar-rudder, line 453: We should generalize this argument. Base class should not have any knowledge of any of its child

src/connectors/SnowflakeConnector.py
```diff
@@ -538,48 +557,106 @@ def get_material_names_(
...
+                    feature_table_name_ = utils.generate_material_name(
+                        material_table_prefix,
```
shekhar-rudder, line 576: Why are we passing `material_table_prefix` as an argument ? The interested functions can directly access it via constants

src/connectors/SnowflakeConnector.py
```diff
@@ -538,48 +557,106 @@ def get_material_names_(
...
                 == prediction_horizon_days,
+                join_type="full",
             ).select(
                 feature_snowpark_df.seq_no.alias("feature_seq_no"),
                 feature_snowpark_df.end_ts.alias("feature_end_ts"),
                 label_snowpark_df.seq_no.alias("label_seq_no"),
                 label_snowpark_df.end_ts.alias("label_end_ts"),
             )
             for row in feature_label_snowpark_df.collect():
+                if (
```
shekhar-rudder, line 568: The whole logic in this for loop is being duplicated in `RedshiftConnector`. We should move up the common logic to `Connector` class

src/connectors/SnowflakeConnector.py
joker2411: after merging both the feature dataframe and label dataframe, it was converting sequence_number to float values. Handled it differently as suggested by @dpatchigolla in new commits.

src/connectors/Connector.py
joker2411: Yes. Both can be not None as well. Corrected it as per the comment.

src/connectors/Connector.py
joker2411: removed this argument in new commits.

src/connectors/SnowflakeConnector.py
joker2411: yes. This is just for better readability, we pass all the values required to create material name in same order.

src/connectors/SnowflakeConnector.py
joker2411: Redshift and Snowflake both vary in accessing their columns and column values. Also, both are using their different sessions to validate the tables. I believe it's better to keep them separate.

src/connectors/Connector.py
```diff
@@ -262,48 +319,59 @@ def validate_historical_materials_hash(
         self,
         session,
         material_table_query: str,
-        feature_material_seq_no: str,
-        label_material_seq_no: str,
+        feature_material_seq_no: int,
```
dpatchigolla, line 322: Can you make typehint Optional[int] to indicate it can be None? same with label_material_seq_no

src/connectors/RedshiftConnector.py
```diff
@@ -445,11 +466,12 @@ def get_material_names_(
                     (df["model_name"] == model_name)
                     & (df["model_type"] == constants.ENTITY_VAR_MODEL)
                     & (df["model_hash"] == model_hash)
-                    & (df["end_ts"] >= start_time)
-                    & (df["end_ts"] <= end_time),
+                    & (df["end_ts"].dt.date >= pd.to_datetime(start_time).date())
```
dpatchigolla, line 469: Can you add unit tests for this? Both in snowflake and redshift connectors? You can mock get_mateiral_registry_table to return a fixed dataframe, where you give predefined end_ts values and test that the logic works as expected?

src/connectors/RedshiftConnector.py
```diff
@@ -445,11 +466,12 @@ def get_material_names_(
...
+                    & (df["end_ts"].dt.date <= pd.to_datetime(end_time).date()),
                     ["seq_no", "end_ts"],
                 ]
                 .drop_duplicates()
+                .assign(end_ts=lambda x: x["end_ts"].dt.date)  # convert to date
```
dpatchigolla, line 474: Why is this required?

src/connectors/RedshiftConnector.py
```diff
@@ -470,46 +492,60 @@ def get_material_names_(
...
                 .rename(columns={"seq_no": "label_seq_no", "end_ts": "label_end_ts"})
             )
 
             feature_label_df = pd.merge(
-                feature_df, label_df, on="label_end_ts", how="inner"
+                feature_df, label_df, on="label_end_ts", how="outer"
             )
+            feature_label_df = feature_label_df.replace({np.nan: None})
```
dpatchigolla, line 507: You can add the `.replace({np.nan: None})` in prev line itself. Variable re-assigning should usually be avoided if possible. So, 
```
feature_label_df = pd.merge(
                feature_df, label_df, on="label_end_ts", how="outer"
            ).replace({np.nan: None})
```

src/connectors/Connector.py
```diff
@@ -262,48 +319,59 @@ def validate_historical_materials_hash(
...
+        label_material_seq_no: int,
     ) -> bool:
         """This function will validate the input material query with seq number from historical material tables
```
dpatchigolla, line 325: _This function will validate the input material query with seq number from historical material tables_

This isn't really very helpful. Can you modify this to following? 

entity_var_table hash doesn't necessarily change if the underlying entity vars change. So we need to validate if the historic entity var tables were generated by the same model that is generating current entity var table. We do that by checking the input tables that generate entity var table, by replacing the current seq no with historic seq nos and verify those tables exist. This relies on the input sql queries sent by profiles which point to the current materialised tables.

src/connectors/RedshiftConnector.py
```diff
@@ -470,46 +492,60 @@ def get_material_names_(
...
-                        )
+                        material_names.append((feature_table_name_, label_table_name_))
                         training_dates.append(
                             (str(row["feature_end_ts"]), str(row["label_end_ts"]))
                         )
+                if (
+                    self._count_fully_defined_lists(material_names)
+                    >= constants.TRAIN_MATERIALS_LIMIT
+                ):
+                    break  # we might be taking more than 10 materials for training but this would consider all types of inputs.
```
dpatchigolla, line 548: >10 materials

Where is 10 coming from? Also I am not sure I understand this comment clearly. Can you try to rephrase it, and add it as a logger.debug instead of as a comment?

src/connectors/SnowflakeConnector.py
```diff
@@ -116,6 +116,27 @@ def get_udf_name(self, model_path: str) -> str:
...
+        Args:
+            session (snowflake.snowpark.Session): A Snowpark session for data warehouse access.
+            table_name (str): The name of the table to be checked.
+
+        Returns:
+            bool: True if the table exists, False otherwise.
+        """
+        try:
+            if table_name is None:
+                return True
```
dpatchigolla, line 134: How is it valid if table_name is None?

dpatchigolla: This, and rest of comments in this is applicable for the RedshiftConnector too - and vice-versa - wherever they are relevant.

dpatchigolla (changes requested):

src/connectors/RedshiftConnector.py
joker2411: removed.

src/connectors/Connector.py
```diff
@@ -192,34 +173,51 @@ def generate_training_materials(
...
-            if material_pair[0] is not None and material_pair[1] is None:
-                feature_table_name_ = material_pair[0]
+        feature_date, label_date = None, None
+        for material_info in materials:
+            if (
+                material_info.feature_table_name is not None
+                and material_info.label_table_name is None
+            ):
+                feature_table_name_ = material_info.feature_table_name
                 if self.is_valid_table(session, feature_table_name_):
```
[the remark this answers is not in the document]
joker2411: Yes, it shouldn't have any invalid tables because entries inside materials are only made after verifying if the table is valid or not.
Raised and assert.

dpatchigolla (approved):

PR #181, opened 2024-02-15
trigger preprocess and train script in local mode

joker2411:
Changes to be committed:
	modified:   src/ml_core/preprocess_and_train.py
	modified:   src/processors/AWSProcessor.py
	modified:   src/processors/LocalProcessor.py
	modified:   src/processors/Processor.py
	modified:   src/processors/SnowflakeProcessor.py

## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/ml_core/preprocess_and_train.py
+++ b/src/ml_core/preprocess_and_train.py
@@ -339,14 +339,13 @@ if __name__ == "__main__":
     with open(os.path.join(local_folder, args.ec2_temp_output_json), "w") as file:
         json.dump(train_results_json, file)
 
-    train_upload_whitelist = [
-        trainer.figure_names["feature-importance-chart"],
-        trainer.figure_names["lift-chart"],
-        trainer.figure_names["pr-auc-curve"],
-        trainer.figure_names["roc-auc-curve"],
-        f"{trainer.output_profiles_ml_model}_{model_file_name}",
-        args.ec2_temp_output_json,
-    ]
+    train_upload_whitelist = utils.merge_lists_to_unique(
+        list(trainer.figure_names.values()),
+        [
+            f"{trainer.output_profiles_ml_model}_{model_file_name}",
+            args.ec2_temp_output_json,
+        ],
+    )
     if args.mode == constants.K8S_MODE:
         logger.debug(f"Uploading trained files to s3://{args.s3_bucket}/{args.s3_path}")
         S3Utils.upload_directory(
--- a/src/ml_core/preprocess_and_train.py
+++ b/src/ml_core/preprocess_and_train.py
@@ -293,6 +293,7 @@ if __name__ == "__main__":
     parser.add_argument("--merged_config", type=json.loads)
     parser.add_argument("--prediction_task", type=str)
     parser.add_argument("--wh_creds", type=json.loads)
+    parser.add_argument("--output_path", type=str)
     parser.add_argument("--mode", type=str)
     args = parser.parse_args()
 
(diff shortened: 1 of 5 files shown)


src/processors/Processor.py
```diff
@@ -23,24 +21,6 @@ def __init__(
         self.connector = connector
         self.session = session
 
-    def train(
```
dpatchigolla, line 26: Why not leave it as an abstractmethod?

src/ml_core/preprocess_and_train.py
```diff
@@ -310,7 +311,11 @@ def preprocess_and_train(
         trainer = RegressionTrainer(**args.merged_config)
 
     # Creating the Redshift connector and session bcoz this case of code will only be triggerred for Redshift
-    current_dir = os.path.dirname(os.path.abspath(__file__))
+    current_dir = (
```
dpatchigolla, line 314: Why is this change required?

src/ml_core/preprocess_and_train.py
```diff
@@ -293,6 +293,7 @@ def preprocess_and_train(
     parser.add_argument("--merged_config", type=json.loads)
     parser.add_argument("--prediction_task", type=str)
     parser.add_argument("--wh_creds", type=json.loads)
+    parser.add_argument("--output_path", type=str)
```
dpatchigolla, line 296: A new parameter got added but it is not used when calling in awsprocessor. how does it work?

src/ml_core/preprocess_and_train.py
```diff
@@ -334,31 +339,32 @@ def preprocess_and_train(
...
             args.s3_bucket,
             args.aws_region_name,
             args.s3_path,
             local_folder,
             train_upload_whitelist,
         )
-    logger.debug(f"Deleting additional local directory from infra mode")
-    connector.cleanup(delete_local_data=True)
+        logger.debug(f"Deleting additional local directory from infra mode")
+        connector.cleanup(delete_local_data=True)
```
dpatchigolla, line 370: Can the logger statements and cleanup command be moved outside the if else blocks?

dpatchigolla: Or can we add mode as a param to upload_directory itself, and merge both functions to one?  This would prevent the need for an if else statment.
Within upload_directory, it attempts to fetch keys. If it finds, it sets up s3 client with those keys. else, without - similar to how it is currently in upload_directory. This may need changes in other places where these functions are currently being called.

dpatchigolla (changes requested):

src/ml_core/preprocess_and_train.py
joker2411: in local processing mode, we'll get the pb_output folder path which we'll use to create connector. But on ec2/k8, that path may or maynot exist because of churn repo not being there. Therefore, on ec2/k8, I kept it relative to project path for temporary use and processing.

src/ml_core/preprocess_and_train.py
joker2411: because in local_mode, output path will be to pb output folder and in case of awsprocessor, we are using current project path. Also, I have checked that if we don't pass any argument, then ```parser.add_argument()``` has a default of None which works fine in our case.

src/ml_core/preprocess_and_train.py
joker2411: Yes, I am also planning to move few things in __main__ into functions itself using mode as an additional parameter. Though I'll be taking that as a separate task after this and don't want to include that in this.

src/processors/Processor.py
joker2411: corrected.

dpatchigolla (approved):

PR #204, opened 2024-02-28
Feature/prml 256 integrate pycaret into classifier

subham-71:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


requirements.txt
```diff
@@ -10,6 +10,7 @@ scikit_learn>=1.1.1
 shap>=0.41.0
 platformdirs>=3.8.1
 xgboost>=1.5.0
+pycaret==3.2.0
```
dpatchigolla, line 13: Can you comment this out? Can't merge till this is in rudder-sources.

samples/application_project/models/profiles-ml.yaml
```diff
@@ -5,7 +5,7 @@ models :
       occurred_at_col: insert_ts
       entity_key: user
       validity_time: 0s # 1 day
-      py_repo_url: ../../../../
+      py_repo_url: ../../
```
dpatchigolla, line 8: Are these changes to samples project for pycaret testing?

samples/predictions_dev_project/pb_project.yaml
```diff
@@ -3,7 +3,7 @@ name: base_features
 # Project's yaml schema version
 schema_version: 44
 # WH Connection Profile to use.
-connection: default
+connection: dev_wh2
```
dpatchigolla, line 6: You can change it back to default

src/ml_core/preprocess_and_predict.py
```diff
@@ -124,7 +124,7 @@ def predict_helper(df, model_name: str, **kwargs) -> Any:
         df[numeric_columns] = df[numeric_columns].replace({pd.NA: np.nan})
         df[categorical_columns] = df[categorical_columns].replace({pd.NA: None})
         if prediction_task == "classification":
-            return trained_model.predict_proba(df)[:, 1]
+            return trained_model.predict(df)
```
dpatchigolla, line 127: Does pycaret return a score in both classification and regression? In that case you don't need if else conditions.

src/trainers/MLTrainer.py
```diff
@@ -452,6 +421,79 @@ def plot_diagnostics(
...
+            test_y,
+            val_x,
+            val_y,
+            train_data,
+            test_data,
+            val_data,
+        ) = self.prepare_data(feature_df, categorical_columns, numeric_columns)
+
+        # Initialize PyCaret setup for classification with train and test data
+        clf_setup = classification_setup(
```
dpatchigolla, line 468: Except for [classification|regression]_setup and [classification|regression]_compare_models, rest all look identical. Is that right? We can do following in that case:
Define this whole function block in the abstract class itself, where you send these two as parameters - 
```
def _train_model(self, ....., setup: Callable, compare_models: Callable):
......
....
setup_ = setup(data=train_data, test_data.....)
best_model = compare_models()
...
```

In ClassificationTrainer, do following
```
def train_model(self, ....):
     self._train_model(....., classification_setup, classification_compare_models)
```

In RegressionTrainer, replace with regression_setup and regression_compare_models functions

Basically, just passing the right functions as paramters to the function, instead of duplicating the code.

src/utils/utils.py
```diff
@@ -195,26 +196,33 @@ def get_metrics_classifier(
...
         metrics = {"train": train_metrics, "val": val_metrics, "test": test_metrics}
-        predictions = {"train": train_preds, "val": val_preds, "test": test_preds}
+        predictions = {
+            "train": train_predictions,
+            "val": val_predictions,
+            "test": test_predictions,
+        }
 
-        return metrics, predictions, round(prob_threshold, 2)
+        return metrics, predictions, round(0, 2)
```
dpatchigolla, line 225: Add a comment next to it 

`#ToDO: This is for compatibility with UI and should be safely removed once the param is removed on UI`

samples/application_project/models/profiles-ml.yaml
subham-71: Yes. Earlier it was picking up the directories correctly , but it started throwing up errors recently. So, I had to change the relative path.

PR #214, opened 2024-03-07
redshift write through S3 fix

joker2411:
Changes to be committed:
	modified:   src/predictions/rudderstack_predictions/connectors/wh/connector_base.py
	modified:   src/predictions/rudderstack_predictions/connectors/wh/redshift_base_connector.py

## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/predictions/rudderstack_predictions/connectors/wh/connector_base.py
+++ b/src/predictions/rudderstack_predictions/connectors/wh/connector_base.py
@@ -5,6 +5,13 @@
 import pandas as pd
 
 
+def get_create_statement(df: pd.DataFrame, table_name: str, schema: str = None) -> None:
+    create_statement = pd.io.sql.get_schema(df, table_name, schema=schema)
+    return create_statement.replace('"', "").replace(
+        "CREATE TABLE", "CREATE TABLE IF NOT EXISTS"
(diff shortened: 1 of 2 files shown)


src/predictions/rudderstack_predictions/connectors/wh/connector_base.py
```diff
@@ -5,6 +5,13 @@
 import pandas as pd
 
 
+def get_create_statement(df: pd.DataFrame, table_name: str, schema: str = None) -> None:
```
shekhar-rudder, line 8: Please add details how this change fixes the bug

joker2411: the except block created to catch the error in redshift_base_connect if the table didn't exist wasn't correct. So, correct that which then triggers the connector_base to create the table by given name.

But in this script, ```get_create_statement()```  was called but defined nowhere. So, checked the ```pywht/profiles_rudderstack``` for this and taken it from there. It doesn't need any additional further packages, so adding only this worked. Checked by uploading a sample table.

PR #221, opened 2024-03-12
refactor: extract out wht specific logic in a separate class

shekhar-rudder:
## Description of the change

- Updated path of `base_features` package in `application_features` project
- Moved out wht specific logic from `Connector` class to a new class `PythonWht`
- Created a new method in the child connector classes `join_feature_label_tables`
- Moved `check_and_generate_more_materials` from `Connector` to `MLTrainer` class
- Got rid of `get_project_folder` logic

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/rudderstack_predictions/connectors/CommonWarehouseConnector.py
```diff
@@ -333,91 +314,61 @@ def fetch_filtered_table(
...
+            start_time,
+            end_time,
+            columns={"seq_no": "FEATURE_SEQ_NO", "end_ts": "FEATURE_END_TS"},
+        )
+        required_feature_cols = feature_df.columns.to_list()
+        feature_df["TEMP_LABEL_END_TS"] = feature_df["FEATURE_END_TS"] + timedelta(
+            days=prediction_horizon_days
+        )
 
-            for _, row in feature_label_df.iterrows():
```
dpatchigolla, line 407: You moved this code to wht abstraction but kept the rest of the logic here because that's not warehouse agnostic?

src/predictions/rudderstack_predictions/wht/pythonWHT.py
```diff
@@ -0,0 +1,457 @@
...
+        A sequence is said to be complete if it does not have any Nones.
+        """
+        complete_sequences = [
+            sequence
+            for sequence in sequences
+            if all(element is not None for element in sequence)
+        ]
+        return complete_sequences
+
+    def _get_material_names_(
```
dpatchigolla, line 179: `_get_material_names` sounds better than this. Can you remove the trailing `_`?

src/predictions/rudderstack_predictions/wht/pythonWHT.py
```diff
@@ -0,0 +1,457 @@
...
+        if validation_flag:
+            train_table_info = TrainTablesInfo(
+                feature_table_name=feature_material_name,
+                feature_table_date=str(table_row.FEATURE_END_TS),
+                label_table_name=label_material_name,
+                label_table_date=str(table_row.LABEL_END_TS),
+            )
+            materials.append(train_table_info)
+
+    def _get_complete_sequences(self, sequences: List[Sequence]) -> int:
```
dpatchigolla, line 168: Suggestion: Does it make sense to move this to inside `get_material_names` where this is being used? As this is a pure utility function and nothing to do with wht?

src/predictions/rudderstack_predictions/wht/pythonWHT.py
```diff
@@ -0,0 +1,457 @@
...
+
+def split_key(item):
+    parts = item.split("_")
+    if len(parts) > 1 and parts[-1].isdigit():
+        return int(parts[-1])
+    return 0
+
+
+class PythonWHT:
+    def init(
```
dpatchigolla, line 23: Is there a reason why you didn't define it as `__init__`? Why not initialise the object directly with required params?

src/predictions/rudderstack_predictions/wht/mockPB.py
dpatchigolla, line 1: Is this still being used now anywhere?

dpatchigolla (approved):
Minor suggestions. Looks good overall. Amazing job at refactoring!!

src/predictions/rudderstack_predictions/wht/mockPB.py
shekhar-rudder: Yes, it is still being used in getPB which is now being used in [PythonWHT](https://github.com/rudderlabs/rudderstack-profiles-classifier/pull/221/files#diff-9db256096909c03511403532b8d09e01d54c5f2b61c51cf93100abedb843b080R36)

src/predictions/rudderstack_predictions/wht/pythonWHT.py
shekhar-rudder: `whtService` needs to be passed down as an argument in train2. So to use `__init__` instead of `init`, we need to create the object in train1. This would also mean creating both session and connector objects as well in train1 which is currently being done in train2. We can probably do that, let me think about it more in the next PR

src/predictions/rudderstack_predictions/wht/pythonWHT.py
shekhar-rudder: `_get_complete_sequences` is being called twice in `get_material_names` with different values of `materials`. So it needs to be extracted as a method. Are you suggesting to have it as a method itself inside `get_material_names` ?

src/predictions/rudderstack_predictions/connectors/CommonWarehouseConnector.py
shekhar-rudder: Yes. The code which was being duplicated in the 2 connectors was moved to wht abstraction. The warehouse specific code remained behind

src/predictions/rudderstack_predictions/wht/pythonWHT.py
shekhar-rudder: Renamed to `_get_material_names`

src/predictions/rudderstack_predictions/wht/pythonWHT.py
dpatchigolla: Got it. May be we can still just pass the class as a parameter in train1, instead of the object, then construct the object in train2. But yes, can be in the next pr.

src/predictions/rudderstack_predictions/wht/pythonWHT.py
dpatchigolla: Yes. a locally scoped method within get_material_names.

shekhar-rudder: Made it locally scoped

PR #227, opened 2024-03-18
feat: implement training pynative model recipe

shekhar-rudder:
## Description of the change

- Added another implementation of the WhtService interface called PyNativeWht. For most of the methods, it is using PythonWht to implement them. The plan is to phase it out incrementally
- Added a new argument `ml_core_path` in the train function since its value is going to be different for Python and PyNative models
- Created a method `get_model_creation_ts` in pythonWht so that it can be used by PyNativeWht
- Removed `_` from `_split_material_name` in pythonWht so that it can be used by PyNativeWht

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/rudderstack_predictions/wht/pyNativeWHT.py
```diff
@@ -0,0 +1,56 @@
...
+        self.whtMaterial = whtMaterial
+
+    def init(
+        self,
+        connector: Connector,
+        session,
+        site_config_path: str,
+        project_folder_path: str,
+    ) -> None:
+        self.pythonWHT.init(connector, session, site_config_path, project_folder_path)
```
dpatchigolla, line 21: From now on, will pythonWHT references be made only in this file?

src/predictions/rudderstack_predictions/processors/ProcessorFactory.py
```diff
@@ -5,17 +5,16 @@
...
 class ProcessorFactory:
-    def create(mode: str, trainer, connector, session):
+    def create(mode: str, trainer, connector, session, ml_core_path: str):
         processor = None
         if mode == constants.RUDDERSTACK_MODE:
             # Lazy load K8sProcessor since kubernetes might not be installed in all environments
             from .K8sProcessor import K8sProcessor
 
-            processor = K8sProcessor
+            return K8sProcessor(trainer, connector, session, ml_core_path)
```
dpatchigolla, line 14: Why not keep it as earlier? Just add the ml_core_path in the single return statement?

src/predictions/rudderstack_predictions/py_native/training.py
```diff
@@ -2,36 +2,113 @@
...
+            "ml_config": {
+                "type": "object",
+                "additionalProperties": False,
+                "properties": {
+                    "data": {
+                        "type": "object",
+                        "additionalProperties": True,
+                        "properties": {
+                            "label_column": {"type": "string"},
+                            "prediction_horizon_days": {"type": "integer"},
```
dpatchigolla, line 32: These are the mandatory fields but there are many other optional params. They will be added later right?

src/predictions/rudderstack_predictions/py_native/training.py
```diff
@@ -2,36 +2,113 @@
...
-        pass
+        import os
+        import yaml
+
+        homedir = os.path.expanduser("~")
+        with open(os.path.join(homedir, ".pb/siteconfig.yaml"), "r") as f:
+            creds = yaml.safe_load(f)["connections"]["test"]["outputs"]["env"]
+        input_model_refs = self.build_spec.get("inputs", [])
+        output_filename = self._get_train_output_filepath(this)
+        site_config_path = os.getenv("SITE_CONFIG_PATH")
```
dpatchigolla, line 95: How and where are the env variables set?
Also site_config_path is used above in line 91. shouldn't that come after this line and use a single site config path?

src/predictions/rudderstack_predictions/trainers/MLTrainer.py
```diff
@@ -315,9 +315,11 @@ def check_and_generate_more_materials(
         materials: List[TrainTablesInfo],
         input_models: str,
         whtService: PythonWHT,
+        connector: Connector,
+        session,
     ):
         met_data_requirement = self.check_min_data_requirement(
-            whtService.connector, whtService.session, materials
```
dpatchigolla, line 320: why is this change required? PythonWHT still has connector and session objects right

src/predictions/rudderstack_predictions/wht/pyNativeWHT.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def get_latest_entity_var_table(self, entity_key: str) -> Tuple[str, str, str]:
```
dpatchigolla, line 23: Can we add unit tests for these pynative functions? May be by mocking the whtmaterial (can we use the mockpb for that?)

dpatchigolla (approved):
Looks good. Have a few suggestions/clarifications.

src/predictions/rudderstack_predictions/wht/pyNativeWHT.py
shekhar-rudder: No, the outermost train.py file will still reference pythonWHT. This is just a way of reusing the existing code in pyNativeWHT

src/predictions/rudderstack_predictions/py_native/training.py
shekhar-rudder: Yes there are many other optional params. I don't have an opinion right now if we should add all of them (because they are too many) but yes we can add the important ones later

src/predictions/rudderstack_predictions/trainers/MLTrainer.py
shekhar-rudder: Yes PythonWHT still has connector and session objects but in the py native flow, MLTrainer will get the pyNativeWHT object. So the current code won't work (it should be whtService.pythonWHT.connector). So decoupled connector and session obects from whtService

src/predictions/rudderstack_predictions/processors/ProcessorFactory.py
shekhar-rudder: Reverted back

PR #245, opened 2024-04-03
Fixed some issues with defaults in smart dataset. See description for details

dpatchigolla:
1. changed constants  for data validation to more reasonable numbers
2. added extra warning messages on materialisation strategy if unnecessary params are being sent
3. separated out materialisation config loading as a function with additional assertions on the format
4. if manual strategy, all dates are materialised irrespective of data validation

## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


cnu1439 (approved):

cnu1439 (approved):

src/predictions/rudderstack_predictions/trainers/MLTrainer.py
```diff
@@ -58,23 +58,59 @@ def __init__(
...
+                    "feature_data_min_date_diff is not required for manual materialisation strategy. It gets ignored."
+                )
+        elif self.materialisation_strategy == "auto":
+            self.materialisation_max_no_dates = int(
+                materialisation_config["max_no_of_dates"]
+            )
+            self.feature_data_min_date_diff = int(
+                materialisation_config["feature_data_min_date_diff"]
+            )
+            assert (
```
joker2411, line 99: I think these assert checks should be before the value assignment right above this. Because right now, we are doing the assignments and then assert checks, so, if they are not present we'll be getting "wrong keyword errors" instead of assert erros. (suggestion for both the asserts in this elif).

scripts/simulate_pb_run.py
```diff
@@ -92,17 +92,17 @@
         "entity_key": entity_key,
```
joker2411, line 92: Just needed bit clarification in this, 
1. since we have removed 'inputs' as one of the parameters from data config(from both profiles-ml.yaml and model_configs.yaml), have you tried after removing it from here and then checked ? Because just wanted to know how it would figure out the required '-m' flag inputs while materialising past features.
2. Also, even if passing it here is correct, inputs used in profiles-ml.yaml are different from the ones created here, So why 2 different inputs ?

src/predictions/rudderstack_predictions/trainers/MLTrainer.py
```diff
@@ -58,23 +58,59 @@ def __init__(
...
+            "manual",
+            "",
+        ], "materialisation strategy can only be 'auto', 'manual', or ''."
+        if self.materialisation_strategy == "manual":
+            self.materialisation_dates = materialisation_config["dates"]
+            assert (
+                len(self.materialisation_dates) > 0
+            ), "materialisation dates are required for manual strategy."
+            if "max_no_of_dates" in materialisation_config:
+                logger.warning(
```
shekhar-rudder, line 85: I don't think there is a value in showing warning for unnecessary fields, makes the code look complex. Either we throw an error or don't have any checks

dpatchigolla: Okay, will remove the check.

scripts/simulate_pb_run.py
dpatchigolla: We are popping the inputs from config, to have backward compatibility for any UI projects. So having it is harmless. It gets ignored. We use the input sql queries and `pb show models --json` to figure out the actual inputs that we need to pass in the subprocess.

PR #271, opened 2024-04-15
Feature/prml 479 better error message for invalid rows in date time

chetana1410:
Modified error messages in validate_label_distinct_values in SnowflakeConnector.py and CommonWarehouseConnector.py.
Added exception block in add_days_diff in CommonWarehouseConnector.py.
Modified unit test test_expects_error_if_label_count_is_low_regression so that it matches with the error message in SnowflakeConnector.py


chetana1410:
Had to make change in redshiftConnector.py unit test as well

PR #274, opened 2024-04-19
Bug/prml 510 bug only one set of table pairs is currently being used for

dpatchigolla:
## Description of the change

Fixed a bug around model not picking mutiple materials data correctly.
Also added a simulate script that can work on the preprocess_and_train part directly, with given inputs

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/scripts/simulate_preprocess_run_locally.py
+++ b/scripts/simulate_preprocess_run_locally.py
@@ -21,8 +21,8 @@ from src.predictions.rudderstack_predictions.connectors.ConnectorFactory import 
 # print current working directory
 print(os.getcwd())
 # The next four vars should be modified based on the run.
-profiles_yaml_path = "../../base_features/models/profiles-ml.yaml"
-connection_name = "test"
+profiles_yaml_path = "samples/predictions_dev_project/models/profiles-ml.yaml"
+connection_name = os.getenv("SITE_CONN_NAME", None)
 # You can get these two vars from `ps -ef | grep preprocess_and_train` command when the parent process (simulate_pb_run or pb run) is running and the code enters training part.
 # Or they can be hardcoded based on the run.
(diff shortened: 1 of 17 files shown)


joker2411 (commented):
Rest all looks fine. Just minor clarification needed.

src/predictions/rudderstack_predictions/connectors/SnowflakeConnector.py
```diff
@@ -651,29 +651,18 @@ def drop_cols(
     def filter_feature_table(
```
joker2411, line 651: Since we are no-longer taking single material, previously main idea behind this function was that in case if we are taking multiple materials then there is a possibility that one/many user may occur multiple times in 'concatenated' feature_table. So, this function was taking only one instance of each user and then checking the MIN-MAX samples to maintain the diversity in data.
Though this function would work as well but there may occur single user multiple times. Is that fine? (though diversity wouldn't be an issue in large/actual data as even multiple entries of any user will be very insignificant to whole feature table.)

dpatchigolla: Yes. With the whole smart-dataset thing, we are taking same user multiple-times by design. We have a min gap between feature dates param, which defaults to 2 weeks and customers can control that. So even same user data will look different enough.

scripts/simulate_preprocess_run_locally.py
```diff
@@ -0,0 +1,111 @@
...
+    TrainerFactory,
+)
+from src.predictions.rudderstack_predictions.connectors.ConnectorFactory import (
+    ConnectorFactory,
+)
+
+# print current working directory
+print(os.getcwd())
+# The next four vars should be modified based on the run.
+profiles_yaml_path = "../../base_features/models/profiles-ml.yaml"
```
joker2411, line 24: I guess we can give `profiles-ml.yaml` path from any of the sample projects or from the `.env` that we have right now, it contains project folder path which we can use. That'll make this variable more static and we wouldn't have to change it.

scripts/simulate_preprocess_run_locally.py
```diff
@@ -0,0 +1,111 @@
...
+connection_name = "test"
```
joker2411, line 25: already have it in .env.

joker2411 (approved):

PR #288, opened 2024-04-29
Added logic for the llm model

RuLeKiLLeR:
## Description of the change

Added logic for the llm model

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -25,9 +26,29 @@ def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> No
...
+            "mistral-large",
+            "reka-flash",
+            "mixtral-8x7b",
+            "llama2-70b-chat",
+            "mistral-7b",
+            "gemma-7b",
+        ]
+        if self.build_spec["llm_model_name"] not in valid_models:
+            raise ValueError(
+                f"Invalid model name: {self.build_spec['llm_model_name']}. Valid options are: {', '.join(valid_models)}"
```
shekhar-rudder, line 43: `Invalid model name` -> This is confusing since profiles also has a concept of model name. It can be worded as `Invalid llm model name`

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -25,9 +26,29 @@ def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> No
...
     def validate(self):
+        valid_models = [
+            "mistral-large",
+            "reka-flash",
+            "mixtral-8x7b",
+            "llama2-70b-chat",
+            "mistral-7b",
+            "gemma-7b",
+        ]
+        if self.build_spec["llm_model_name"] not in valid_models:
```
shekhar-rudder, line 41: There should be a unit test which tests the following 3 scenarios
1. `llm_model_name` is None
2. `llm_model_name` is `mistral-large`
3. `llm_model_name` is `random-model`

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -25,9 +26,29 @@ def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> No
...
+            raise ValueError(
+                f"Invalid model name: {self.build_spec['llm_model_name']}. Valid options are: {', '.join(valid_models)}"
+            )
+
+        prompt = self.build_spec["prompt"]
+        input_count = prompt.count("{")
+        input_lst = self.build_spec["prompt_inputs"]
+        prompt_input_count = len(input_lst)
+        if input_count != prompt_input_count:
```
shekhar-rudder, line 50: Unit tests to cover following scenarios
1. No "{" in the input string
2. `input_count` and `prompt_input_count` do not match
3. `prompt_input_count` and `input_count` match

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -25,9 +26,29 @@ def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> No
...
+            "mistral-7b",
+            "gemma-7b",
+        ]
+        if self.build_spec["llm_model_name"] not in valid_models:
+            raise ValueError(
+                f"Invalid model name: {self.build_spec['llm_model_name']}. Valid options are: {', '.join(valid_models)}"
+            )
+
+        prompt = self.build_spec["prompt"]
+        input_count = prompt.count("{")
```
shekhar-rudder, line 47: - What happens if there are no "}" in the prompt ?
- Is `{}` a valid input ?

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +64,49 @@ def describe(self, this: WhtMaterial):
...
+        )  # This is to escape single quotes
+        input_indices = re.findall(r"{(\w+)\[(\d+)\]}", prompt_replaced)
+        for word, index in input_indices:
+            index = int(index)
+            if 0 <= index < len(input_columns):
+                placeholder = f"{{{word}[{index}]}}"
+                col = "' ||" + input_columns[index] + "|| ' "
+                prompt_replaced = prompt_replaced.replace(placeholder, col)
+            else:
+                self.logger.error(f"Index {index} out of range for input_columns list.")
```
shekhar-rudder, line 87: Any reason why we are not throwing error ?

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +64,49 @@ def describe(self, this: WhtMaterial):
         return self.sql, ".sql"
 
     def prepare(self, this: WhtMaterial):
-        self.sql = ""
+        entity = this.model.entity()
+        name = this.model.name()
+        if entity is None:
+            raise Exception("no entity found for model")
+        input_columns = [
+            f"{{{{{var}.Model.Name()}}}}"  # Assuming var is already in the format "user.Var('profession')"
```
shekhar-rudder, line 72: Why are there so many brackets `{{{{` ?

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +64,49 @@ def describe(self, this: WhtMaterial):
...
+            for var in self.prompt_inputs
+        ]
+
+        prompt_replaced = self.prompt.replace(
```
shekhar-rudder, line 76: `prompt_replaced` computation should have unit tests

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +64,49 @@ def describe(self, this: WhtMaterial):
...
+            index = int(index)
+            if 0 <= index < len(input_columns):
+                placeholder = f"{{{word}[{index}]}}"
+                col = "' ||" + input_columns[index] + "|| ' "
+                prompt_replaced = prompt_replaced.replace(placeholder, col)
+            else:
+                self.logger.error(f"Index {index} out of range for input_columns list.")
+
+        # get the var table from a var
+        # var_table_ref = f"this.DeRef(makePath({entity_key}.Var('{self.vars[0]}').Model.GetVarTableRef()))"
```
shekhar-rudder, line 90: This comment can be removed

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +64,49 @@ def describe(self, this: WhtMaterial):
         return self.sql, ".sql"
 
     def prepare(self, this: WhtMaterial):
-        self.sql = ""
+        entity = this.model.entity()
+        name = this.model.name()
```
shekhar-rudder, line 68: `name` -> `column_name`

src/predictions/rudderstack_predictions/py_native/llm.py
shekhar-rudder, line 72: Instead of using the `.Name()` method, we should be using `.DbObjectNamePrefix()`

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +64,49 @@ def describe(self, this: WhtMaterial):
...
+            f"this.DeRef(makePath({self.prompt_inputs[0]}.Model.GetVarTableRef()))"
+        )
+
+        # model_creator_sql
+        query_template = f"""
+            {{% macro begin_block() %}}
+                {{% macro selector_sql() %}}
+                    {{% set entityVarTable = {var_table_ref} %}}
+
+                    SELECT *, SNOWFLAKE.CORTEX.COMPLETE('{self.llm_model_name}','{prompt_replaced}') AS {name} FROM {{{{entityVarTable}}}}
```
shekhar-rudder, line 101: Why are we copying all the columns from `entityVarTable` ? We should get only the identifier column

shekhar-rudder, line 101: Why so many braces -> `{{{{entityVarTable}}}}`

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +64,49 @@ def describe(self, this: WhtMaterial):
...
+                {{% endmacro %}}
+                {{% exec %}} {{{{warehouse.CreateReplaceTableAs(this.Name(), selector_sql())}}}} {{% endexec %}}
+            {{% endmacro %}}
+
+            {{% exec %}} {{{{warehouse.BeginEndBlock(begin_block())}}}} {{% endexec %}}"""
+
+        self.sql = this.execute_text_template(query_template)
```
shekhar-rudder, line 108: Can you share the sql query which is getting generated corresponding to this model in the `output` folder ?

src/predictions/rudderstack_predictions/py_native/llm.py
RuLeKiLLeR: No along with {} we also need an indices so that we can pinpoint the array element that we need to put there.{[0]} is a valid input

shekhar-rudder: Where are we checking for array element inside braces ?

RuLeKiLLeR: Inside the prepare function

src/predictions/rudderstack_predictions/py_native/llm.py
RuLeKiLLeR: this is format provided to me by shubham I tried running the code without these braces and it is showing me and error, I will ask him and let you know.

src/predictions/rudderstack_predictions/py_native/llm.py
RuLeKiLLeR: Will add an exception here as well.

src/predictions/rudderstack_predictions/py_native/llm.py
RuLeKiLLeR: CREATE OR REPLACE TABLE "MATERIAL_SF_CORTEX_TEST_CA452A02_24" AS (
                    SELECT *, SNOWFLAKE.CORTEX.COMPLETE('mistral-7b','The input field is a job title entered by a person. Map it to a single value from the following list of standardised job titles: [Software Engineer, Data Analyst, ML Engineer, Solutions Architect, Others]. Return only standardised titles in maximum three words ' ||profession|| ' ') AS gender FROM CORTEX_TEST.MATERIAL_USER_VAR_TABLE_A79AF7D6_24
                );   
	END;

src/predictions/rudderstack_predictions/py_native/llm.py
RuLeKiLLeR: I think we first concatenated the custom table columns with the table that is made by profiles. for this i copied some user ID's from profiles tables, then from the concatenated table I am finally fetching the data.

RuLeKiLLeR: Don't know the specific reason but removing even 1 pair is showing syntax error.

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -25,9 +26,29 @@ def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> No
         super().__init__(build_spec, schema_version, pb_version)
 
     def get_material_recipe(self) -> PyNativeRecipe:
+        DEFAULT_LLM_MODEL = "llama2-70b-chat"
```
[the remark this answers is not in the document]
shubhammehra4: We should be using this, instead of passing the entire build_spec pass the values Recipe expects.

src/predictions/rudderstack_predictions/py_native/llm.py
shubhammehra4: @shekhar-rudder We are using python f strings { needs to be escaped in the f string so we have to have double the number of expected {

src/predictions/rudderstack_predictions/py_native/llm.py
shubhammehra4: better to create a function and move the logic inside that, easier for testing

src/predictions/rudderstack_predictions/py_native/llm.py
shubhammehra4: @shekhar-rudder Same, Python f strings

src/predictions/rudderstack_predictions/py_native/llm.py
shubhammehra4, line 41:
```suggestion
        if self.build_spec.get("llm_model_name", DEFAULT_LLM_MODEL) not in valid_models:
```

src/predictions/rudderstack_predictions/py_native/llm.py
shekhar-rudder: > for this i copied some user ID's from profiles tables
You are not copying the user Id column. `SELECT *` will select all the columns in the user table which we don't want

src/predictions/rudderstack_predictions/py_native/llm.py
shekhar-rudder: Why not have all the validations at a single place ?

RuLeKiLLeR: No for the array elements i need to change the array elements positions for column names, So actually I am not checking the array elements in the prepare function but I am replacing those with the column names in the prepare function.

src/predictions/rudderstack_predictions/py_native/llm.py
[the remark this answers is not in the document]
RuLeKiLLeR: Re I am using in a regix in the prepare function.Line no. 78

src/predictions/rudderstack_predictions/py_native/llm.py
RuLeKiLLeR: @shubhammehra4 Can you clear this?

src/predictions/rudderstack_predictions/py_native/llm.py
shekhar-rudder: I'm fine with replacing column names in the prepare function. But if possible, we should not have any case where a validation logic is present in "prepare" but not in "validate". The idea is to give feedback to the user as soon as possible via the "validate" method instead of them waiting for "prepare" to be called. If you are worried about code code duplication between the two methods, then the logic can be extracted out in a function and used by both.

src/predictions/rudderstack_predictions/py_native/llm.py
shubhammehra4: Makes sense, Instead of doing `SELECT *` which will copy all vars, we just need to copy the vars required. We can have `SELECT {input_columns.join(", "}, SNOWFLAKE.CORTEX.COMPLETE('{self.llm_model_name}','{prompt_replaced}') AS {name} FROM {{{{entityVarTable}}}}`

src/predictions/rudderstack_predictions/py_native/llm.py
RuLeKiLLeR: Ok So I think the validation test for {} will have to be removed because there can be a case where the user only wants to use 2 column but has given 3 columns in the input as he will be telling us the index of the column that he wants to use. Let me change the validation to check for the maximum index number that is mentioned in the prompt. If that number is greater then prompt_inputs-1 then we will raise an error.
In the prepare function we won't have to check for that anymore hence both portions will be solved with only single validation the validate function

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +66,44 @@ def describe(self, this: WhtMaterial):
...
-        self.sql = ""
+        entity = this.model.entity()
+        column_name = this.model.name()
+        if entity is None:
+            raise Exception("no entity found for model")
+        input_columns = [
+            f"{{{{{var}.Model.DbObjectNamePrefix()}}}}"  # Assuming var is already in the format "user.Var('profession')"
+            for var in self.prompt_inputs
+        ]
+        joined_columns = ", ".join(input_columns)
```
shekhar-rudder, line 77: No need to copy all the columns, we just want the user id column. I believe the way to get it is `entity_id_column_name = this.model.entity().get("IdColumnName")`

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -28,6 +29,32 @@ def get_material_recipe(self) -> PyNativeRecipe:
         return LLMModelRecipe(self.build_spec)
 
     def validate(self):
+
+        if self.build_spec["llm_model_name"] == "":
```
shekhar-rudder, line 33: Default value assignment should be done in the constructor (init method)

src/predictions/rudderstack_predictions/py_native/llm.py
shubhammehra4: Yes, the statement will be `SELECT {entity_id_column_name},  SNOWFLAKE.CORTEX.COMPLETE(...) AS {name} FROM {{{{entityVarTable}}}}`

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -28,11 +29,36 @@ def get_material_recipe(self) -> PyNativeRecipe:
...
+        if max_index >= len(input_lst):
+            raise ValueError(
+                f"Maximum index {max_index} is out of range for input_columns list."
+            )
         return super().validate()
 
 
 class LLMModelRecipe(PyNativeRecipe):
     def __init__(self, build_spec: dict) -> None:
+        if build_spec["llm_model_name"] == "":
```
shekhar-rudder, line 60: Not the constructor of `LLMModelRecipe`. Should be part of constructor of `LLMModel`

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -23,11 +24,35 @@ class LLMModel(BaseModelType):
 
     def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> None:
         super().__init__(build_spec, schema_version, pb_version)
+        if build_spec["llm_model_name"] == "":
```
shubhammehra4, line 27: You are making the assumption if no value is provided it would be an empty string, which is not correct.

```suggestion
        if build_spec.get("llm_model_name", None) is None:
```

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -43,9 +68,44 @@ def describe(self, this: WhtMaterial):
...
+        entity = this.model.entity()
+        column_name = this.model.name()
+        if entity is None:
+            raise Exception("no entity found for model")
+        input_columns = [
+            f"{{{{{var}.Model.DbObjectNamePrefix()}}}}"  # Assuming var is already in the format "user.Var('profession')"
+            for var in self.prompt_inputs
+        ]
+        entity_id_column_name = this.model.entity().get("IdColumnName")
+        prompt_replaced = self.prompt.replace(
```
shubhammehra4, line 80: Do we plan on adding a unit test in a separate PR?

RuLeKiLLeR: yes

src/predictions/rudderstack_predictions/py_native/llm.py
RuLeKiLLeR: Umm then if the user does not put anything in the llm_model_name then it will itself through an error.

RuLeKiLLeR: But I tried it and it shows me this error.
2024-05-06T10:25:21.515+0530    ERROR   compile/compile.go:55   creating project api: loading project: could not create project: failed to read project yaml 
Error: validating project sample_attribution: getting models for folder: models: error listing models: error parsing model build spec for sf_cortex_test: build spec is invalid: [llm_model_name: Invalid type. Expected: string, given: null]

RuLeKiLLeR: ok then I will remove the empty string check

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -28,6 +29,28 @@ def get_material_recipe(self) -> PyNativeRecipe:
         return LLMModelRecipe(self.build_spec)
 
     def validate(self):
+        valid_model_inputs = [
```
shekhar-rudder, line 32: We need to assign a default value to model name if it is missing from the model spec. If the `BuildSpecSchema` is throwing an error in case of missing model name, then the schema should be updated to allow for missing field

RuLeKiLLeR: hi shashank just went through 3 models in the pycorelib repo, but could not find how to allow for missing fields.

RuLeKiLLeR: Let me search more if I can find it I will commit the new changes

shekhar-rudder: You can check the docs of JSON Schema on how to allow for multiple types
https://cswr.github.io/JsonSchema/spec/multiple_types/

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -15,19 +16,46 @@ class LLMModel(BaseModelType):
...
         "additionalProperties": False,
     }
 
     def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> None:
         super().__init__(build_spec, schema_version, pb_version)
+        if (
+            self.build_spec["llm_model_name"] == ""
+            or self.build_spec["llm_model_name"] == None
+        ):
+            self.build_spec["llm_model_name"] = "llama2-70b-chat"
```
shekhar-rudder, line 31: `llama2-70b-chat` string is being duplicated. We can declare a constant `DEFAULT_LLM_MODEL_NAME` and use that at both the places

RuLeKiLLeR: changed it

src/predictions/rudderstack_predictions/py_native/llm.py
```diff
@@ -15,19 +16,46 @@ class LLMModel(BaseModelType):
...
+            "llm_model_name": {"type": ["string", "null"]},
         },
         "required": ["prompt", "prompt_inputs"] + EntityKeyBuildSpecSchema["required"],
         "additionalProperties": False,
     }
 
     def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> None:
         super().__init__(build_spec, schema_version, pb_version)
+        if (
+            self.build_spec["llm_model_name"] == ""
```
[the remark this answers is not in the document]
shekhar-rudder: This comment also needs to be addressed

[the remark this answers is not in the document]
RuLeKiLLeR: done

shekhar-rudder (approved):

PR #289, opened 2024-05-01
Feature/prml 454 arraytype transformation should have a max categories

subham-71:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/rudderstack_predictions/connectors/CommonWarehouseConnector.py
```diff
@@ -40,7 +40,7 @@ def call_procedure(self, *args, **kwargs):
         return train_function(*args, **kwargs)
 
     def transform_arraytype_features(
-        self, feature_df: pd.DataFrame, arraytype_features: List[str]
+        self, feature_df: pd.DataFrame, arraytype_features: List[str], top_k_columns=2
```
dpatchigolla, line 43: Can you make the default 5? 
And add this in model_config preprocess param so it can be controlled?

tests/unit/RedshiftConnector.py
```diff
@@ -1060,27 +1065,16 @@ def test_df_with_array_features(self):
...
                 "COL3_P1",
-                "COL3_P2",
-                "COL3_P3",
                 "COL3_P4",
-                "COL4_A1",
+                "COL3_OTHERS",
                 "COL4_A2",
                 "COL4_A3",
-                "COL4_A4",
+                "COL4_OTHERS",
```
dpatchigolla, line 1072: How does a tie break happen?
Can you add another test where you have 3 distinct values, all equal, but k=2?
Which one gets filtered out?

dpatchigolla: Also add another test where distinct values <= k

subham-71: > How does a tie break happen? Can you add another test where you have 3 distinct values, all equal, but k=2? Which one gets filtered out?

For tie break , it picks up the first encountered value.

tests/integration/requirements.txt
```diff
@@ -11,7 +11,7 @@ scikit_learn==1.3.2
 shap==0.44.0
 platformdirs==3.11.0
 xgboost==2.0.3
-profiles-rudderstack==0.11.0
+profiles-rudderstack==0.13.1
```
dpatchigolla, line 14: Do we need profiles-rudderstack 0.13.1 for this change? Or this is a generic addition?

dpatchigolla (approved):

tests/integration/requirements.txt
subham-71: This is to pass the integration test since the project was updated

PR #291, opened 2024-05-01
added fbeta score

subham-71:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/predictions/rudderstack_predictions/trainers/ClassificationTrainer.py
+++ b/src/predictions/rudderstack_predictions/trainers/ClassificationTrainer.py
@@ -110,15 +110,10 @@ class ClassificationTrainer(MLTrainer):
             models_to_include,
         )
 
-    def get_metrics(self, model, X_test, y_test, y_train, fold_param) -> dict:
-        train_metrics = (
-            classification_results_pull().loc[("CV-Train", "Mean")].to_dict()
-        )
-        val_metrics = classification_results_pull().loc[("CV-Val", "Mean")].to_dict()
+    def get_metrics(self, model, X_test, y_test, X_train, y_train, fold_param) -> dict:
+        train_metrics = self._evaluate_classifier(model, X_train, y_train)
         test_metrics = self._evaluate_classifier(model, X_test, y_test)
-
-        train_metrics = self.map_metrics_keys(train_metrics)
-        val_metrics = self.map_metrics_keys(val_metrics)
+        val_metrics = train_metrics
 
         val_metrics["users"] = int(1 / fold_param * len(y_train))
         train_metrics["users"] = len(y_train) - val_metrics["users"]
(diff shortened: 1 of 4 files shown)


src/predictions/rudderstack_predictions/trainers/ClassificationTrainer.py
```diff
@@ -163,7 +170,7 @@ def get_classification_metrics(
         }
         return metrics
 
-    def get_metrics_classifier(
+    def _get_metrics_classifier(
```
dpatchigolla, line 173: I didn't realise both are called within the same class. `_get_classification_metrics` and `_get_metrics_classifier` don't sound good. Can you rename one of these? I think we can rename `_get_metrics_classifier` to `_evaluate_classifier`. So:

1. `_get_classification_metrics`: This is a generic function on a given actual and predicted y
2. `_evaluate_classifier`: This is on the specific input model where we generate the predictions and are evaluating that classifier.

src/predictions/rudderstack_predictions/trainers/ClassificationTrainer.py
```diff
@@ -87,9 +88,15 @@ def train_model(
                 "name": "roc_auc",
                 "function": roc_auc_score,
                 "greater_is_better": True,
-            }
+            },
+            {
+                "id": "fbeta_score",
+                "name": "fbeta_score",
+                "function": fbeta_score,
```
dpatchigolla, line 95: In the input config, we take the beta score param right, where is that being passed to the model here?

subham-71: The custom metrics are being passed to the model, and the metric will be added during setup. Also the metric_to_optimize will be used to sort the models.

dpatchigolla: How does the code read this [param](https://github.com/rudderlabs/rudderstack-profiles-classifier/blob/699f8036f45ca42c67b91e77687f26cffe9c6a65/src/predictions/rudderstack_predictions/config/model_configs.yaml#L21)?

subham-71: Not utilising this currently, Using the fbeta_score function from sklearn library.

subham-71: Do you want to utilize this parameter here [link](https://github.com/rudderlabs/rudderstack-profiles-classifier/blob/8ed3eeb08cdab1dfa6772f62890005ebfbe08d88/src/predictions/rudderstack_predictions/trainers/ClassificationTrainer.py#L155) ?

dpatchigolla: `fbeta_score` is the function, but this needs a param - the beta, which a user should be able to configure. Without it reading the param from the config, this is just equivalent to f1-score (with beta defaulting to 1)

dpatchigolla (approved):

PR #295, opened 2024-05-08
 Added unit test for pynative llm model

RuLeKiLLeR:
## Description of the change

 Added unit test for pynative llm model

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/tests/unit/llm.py
+++ b/tests/unit/llm.py
@@ -1,59 +1,57 @@
 import unittest
-from unittest.mock import patch
 from src.predictions.rudderstack_predictions.py_native.llm import LLMModel
 
 
 class TestLLMModelValidation(unittest.TestCase):
-    build_spec = {
-        "prompt": "sample prompt input{[0]}",
-        "prompt_inputs": ["input1", "input2"],
-        "llm_model_name": None,
-    }
-    schema_ver = 53
-    pb_version = "v0.11.2"
-
-    @patch("src.predictions.rudderstack_predictions.py_native.llm.re")
-    def test_llm_model_name(self, mock_re):
-        # Defining Schema
-        build_spec = {
+    def setUp(self):
+        self.build_spec = {
             "prompt": "sample prompt input{[0]}",
             "prompt_inputs": ["input1", "input2"],
-            "llm_model_name": None,
+            "llm_model_name": None
         }
+        self.schema_ver = 53
+        self.pb_version = "v0.11.2"
+
+    def test_llm_model_name(self):
+        # Defining Schema
         # Testing empty model name
-        llm_model = LLMModel(build_spec, self.schema_ver, self.pb_version)
-        mock_re.findall.return_value = [("input1", "0"), ("input2", "1")]
+        llm_model = LLMModel(self.build_spec, self.schema_ver, self.pb_version)
         llm_model.validate()
 
         default_llm_model_name = "llama2-70b-chat"
-        self.assertEqual(default_llm_model_name, build_spec["llm_model_name"])
+        self.assertEqual(default_llm_model_name, self.build_spec["llm_model_name"])
+
         # Testing invalid model name
-        build_spec["llm_model_name"] = "invalid_model"
-        llm_model = LLMModel(build_spec, self.schema_ver, self.pb_version)
+        self.build_spec['llm_model_name'] = "invalid_model"
+        llm_model = LLMModel(self.build_spec, self.schema_ver, self.pb_version)
         with self.assertRaises(ValueError):
             llm_model.validate()
 
-    @patch("src.predictions.rudderstack_predictions.py_native.llm.re")
-    def test_max_index_prompt_inputs(self, mock_re):
-        # Mocking LLMModel class
-        llm_model = LLMModel(self.build_spec, 53, "v0.11.2")
-        input_indices = mock_re.findall.return_value = [
-            ("input1", "0"),
-            ("input2", "1"),
-        ]
-        max_index_direct = len(self.build_spec["prompt_inputs"]) - 1
-        # Validate the SnowflakeCortexModel
+    def test_max_index_prompt_inputs(self):
+        # Testing max index prompt inputs
+        # First, test with valid indices
+        self.build_spec['prompt'] = "sample prompt"
+        llm_model = LLMModel(self.build_spec, self.schema_ver, self.pb_version)
         llm_model.validate()
 
-        # Get the max index calculated by SnowflakeCortexModel
-        max_index_snowflake = max(int(index) for _, index in input_indices)
-        are_equal = max_index_direct == max_index_snowflake
+        # Testing the case for correct maximum index
+        self.build_spec['prompt'] = "sample prompt {input1[0]} {input2[1]}"
+        llm_model = LLMModel(self.build_spec, self.schema_ver, self.pb_version)
+        llm_model.validate()
 
-        # Assert
-        self.assertTrue(are_equal)
-        # Testing the case when prompt maximum index is greater then the size of prompt_inputs
-        self.build_spec["prompt"] = "sample prompt {input1[2]} {input2[3]}"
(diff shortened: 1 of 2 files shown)


tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
+import unittest
+from unittest.mock import patch
+from src.predictions.rudderstack_predictions.py_native.llm import LLMModel
+
+
+class TestLLMModelValidation(unittest.TestCase):
+    build_spec = {
```
shekhar-rudder, line 7: Since this `build_spec` is being used just by one test, it should be moved there

tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
...
+        mock_re.findall.return_value = [("input1", "0"), ("input2", "1")]
+        llm_model.validate()
+
+        default_llm_model_name = "llama2-70b-chat"
+        self.assertEqual(default_llm_model_name, build_spec["llm_model_name"])
+        # Testing invalid model name
+        build_spec["llm_model_name"] = "invalid_model"
+        llm_model = LLMModel(build_spec, self.schema_ver, self.pb_version)
+        with self.assertRaises(ValueError):
+            llm_model.validate()
```
shekhar-rudder, line 34: Lets add another assertion on the error message

tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
...
+class TestLLMModelValidation(unittest.TestCase):
+    build_spec = {
+        "prompt": "sample prompt input{[0]}",
+        "prompt_inputs": ["input1", "input2"],
+        "llm_model_name": None,
+    }
+    schema_ver = 53
+    pb_version = "v0.11.2"
+
+    @patch("src.predictions.rudderstack_predictions.py_native.llm.re")
```
shekhar-rudder, line 15: There is no need to patch`re` package. This test can be written without the patch. Patch only when you are having a hard time writing the test without it

tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
...
+        # Testing invalid model name
+        build_spec["llm_model_name"] = "invalid_model"
+        llm_model = LLMModel(build_spec, self.schema_ver, self.pb_version)
+        with self.assertRaises(ValueError):
+            llm_model.validate()
+
+    @patch("src.predictions.rudderstack_predictions.py_native.llm.re")
+    def test_max_index_prompt_inputs(self, mock_re):
+        # Mocking LLMModel class
+        llm_model = LLMModel(self.build_spec, 53, "v0.11.2")
```
shekhar-rudder, line 39: self.schema_ver, self.pb_version

tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
...
+        input_indices = mock_re.findall.return_value = [
+            ("input1", "0"),
+            ("input2", "1"),
+        ]
+        max_index_direct = len(self.build_spec["prompt_inputs"]) - 1
+        # Validate the SnowflakeCortexModel
```
shekhar-rudder, line 45: We should rename `SnowflakeCortexModel`

tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
...
+
+        default_llm_model_name = "llama2-70b-chat"
+        self.assertEqual(default_llm_model_name, build_spec["llm_model_name"])
+        # Testing invalid model name
+        build_spec["llm_model_name"] = "invalid_model"
+        llm_model = LLMModel(build_spec, self.schema_ver, self.pb_version)
+        with self.assertRaises(ValueError):
+            llm_model.validate()
+
+    @patch("src.predictions.rudderstack_predictions.py_native.llm.re")
```
shekhar-rudder, line 36: Same comment here. It should not be patched

tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
...
+        max_index_direct = len(self.build_spec["prompt_inputs"]) - 1
+        # Validate the SnowflakeCortexModel
+        llm_model.validate()
+
+        # Get the max index calculated by SnowflakeCortexModel
+        max_index_snowflake = max(int(index) for _, index in input_indices)
+        are_equal = max_index_direct == max_index_snowflake
+
+        # Assert
+        self.assertTrue(are_equal)
```
shekhar-rudder, line 53: There is no value being added with this particular assertion since we are testing the code by duplicating the logic in the test. This makes the test brittle since the code can change its internal logic and we still be testing the old logic

tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
...
+        # Testing the case when prompt maximum index is greater then the size of prompt_inputs
+        self.build_spec["prompt"] = "sample prompt {input1[2]} {input2[3]}"
+        llm_model = LLMModel(self.build_spec, 53, "v0.11.2")
+        mock_re.findall.return_value = [("input1", "2"), ("input2", "3")]
+        with self.assertRaises(ValueError):
+            llm_model.validate()
```
shekhar-rudder, line 59: there should be another assertion on error message

tests/unit/llm.py
```diff
@@ -0,0 +1,59 @@
...
+        max_index_snowflake = max(int(index) for _, index in input_indices)
+        are_equal = max_index_direct == max_index_snowflake
+
+        # Assert
+        self.assertTrue(are_equal)
+        # Testing the case when prompt maximum index is greater then the size of prompt_inputs
+        self.build_spec["prompt"] = "sample prompt {input1[2]} {input2[3]}"
+        llm_model = LLMModel(self.build_spec, 53, "v0.11.2")
+        mock_re.findall.return_value = [("input1", "2"), ("input2", "3")]
+        with self.assertRaises(ValueError):
```
shekhar-rudder, line 58: Lets test another scenario where the prompt has 0 inputs

tests/unit/llm.py
RuLeKiLLeR: where should it be moved can you specify?

tests/unit/llm.py
RuLeKiLLeR: Shashank can we meet for this as I was getting multiple error messages when I was not patching the re module.

tests/unit/llm.py
RuLeKiLLeR: yes, can we meet up today to fix this, Like I am not getting in what way should i change this. But I got the point.

tests/unit/llm.py
RuLeKiLLeR: but in that case we can't raise an error right. cause the user can give us lesser number of inputs then he has provided in the prompt_inputs. Also in the code too, I am not raising any error when prompt has 0 inputs mentioned.

tests/unit/llm.py
shekhar-rudder: It should be moved inside `test_max_index_prompt_inputs`

tests/unit/llm.py
shekhar-rudder: Sure we can connect. Can you also share what error messages you are getting ?

tests/unit/llm.py
shekhar-rudder: We don't need to raise any errors. The test just makes sure that the code works fine (no need to throw any error) for this particular case

tests/unit/llm.py
RuLeKiLLeR: I think its fixed now gonna commit the new changes. I have added the 0 input case as well.

tests/unit/llm.py
shekhar-rudder: All the places where we are expecting `validate` method to throw an error, we should also validate the content of the error message

RuLeKiLLeR: ok So I have to assert the error also which I am raising right?

shekhar-rudder: Yes

shekhar-rudder (approved):

PR #304, opened 2024-05-16
Removing past materials with similar feature dates.

cnu1439:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/rudderstack_predictions/wht/pythonWHT.py
```diff
@@ -387,6 +389,42 @@ def split_material_name(self, name: str) -> dict:
...
+        materials: List[TrainTablesInfo],
+        prediction_horizon_days: int,
+        feature_data_min_date_diff,
+    ):
+        valid_feature_dates = []
+        valid_materials = []
+
+        for material in materials:
+            feature_date = None
+            if material.feature_table_date != "None":
```
dpatchigolla, line 403: Is it not `None`? Where do we make it a string "None"?

tests/unit/pythonWHT.py
```diff
@@ -162,3 +163,30 @@ def test_missing_sequence_number(self, mock_compute_material_name):
...
+        ]
+        self.pythonWHT = PythonWHT()
+
+    def test_valid_materials(self):
+        valid_materials = self.pythonWHT.get_past_materials_with_valid_date_range(
+            self.materials, 7, 3
+        )
+        self.assertEqual(len(valid_materials), 2)
+        self.assertEqual(valid_materials[0].feature_table_name, "feature_table_1")
+        self.assertEqual(valid_materials[1].feature_table_name, "feature_table_3")
```
dpatchigolla, line 186: How are we getting feature_table_3 too? 16 + 3 would be 19, so this shouldn't come in right?

src/predictions/rudderstack_predictions/wht/pythonWHT.py
```diff
@@ -387,6 +389,42 @@ def split_material_name(self, name: str) -> dict:
...
+    def get_past_materials_with_valid_date_range(
+        self,
+        materials: List[TrainTablesInfo],
+        prediction_horizon_days: int,
+        feature_data_min_date_diff,
+    ):
+        valid_feature_dates = []
+        valid_materials = []
+
+        for material in materials:
```
dpatchigolla, line 401: Can you sort the materials first, so we bias towards recent material? Ex: 
Among feature dates `2024-05-01` and `2024-05-02`, we should pick 2nd May.

dpatchigolla: Can you also add a test case to ensure this happens?

dpatchigolla (changes requested):

src/predictions/rudderstack_predictions/wht/pythonWHT.py
cnu1439: "None" is being set in `pythonWHT.py` line 218 to 224

tests/unit/pythonWHT.py
cnu1439: I used existing logic to find the valid date ( see `utils.py` +458). Considering the dates when the difference is equals to valid difference. If that is not the intended behavior, I will update it to not to consider if it equals.

src/predictions/rudderstack_predictions/wht/pythonWHT.py
cnu1439: Its little complex to sort the materials because some dates will have "None" values. since we are getting materials dates form the warehouse sorted by `creation_ts`. I thought I could skip the sorting here.

tests/unit/pythonWHT.py
dpatchigolla: We are using `<`, not `<=`
This is fine.

dpatchigolla (approved):

PR #329, opened 2024-05-23
Necessary changes for adding the predicted column by snowflake cortex to customer 360 table

RuLeKiLLeR:
## Description of the change

add llm prediction output columns as features and ensure

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


samples/llm_model/models/profiles.yaml
```diff
@@ -24,6 +30,12 @@ models:
       prompt_inputs:
         - user.Var('profession')
       llm_model_name: "mistral-7b"
+      features:
+        - name: llm_profession_prediction
+      ids:
```
shekhar-rudder, line 35: ids need not be added in this yaml file

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -80,7 +87,7 @@ def __init__(self, build_spec: dict) -> None:
     def describe(self, this: WhtMaterial):
         return self.sql, ".sql"
 
-    def prepare(self, this: WhtMaterial):
+    def register_dependencies(self, this: WhtMaterial):
```
shekhar-rudder, line 90: Why are we renaming this function ?

shekhar-rudder (commented):
Please fix the PR title

samples/llm_model/models/profiles.yaml
RuLeKiLLeR: @shubhammehra4 do let me know if you want to add something here

src/predictions/profiles_mlcorelib/py_native/llm.py
RuLeKiLLeR: @shubhammehra4 told me to migrate from prepare to register_dependencies

shubhammehra4, line 90: The model creator sql is invalid.
# is not valid for comments in sql, use this instead.

Are the comments required in SQL?

RuLeKiLLeR: yeah let me fix it in this PR too

RuLeKiLLeR: They are needed as the sql is going to become more complex

samples/llm_model/models/profiles.yaml
RuLeKiLLeR: @shubhammehra4 Can you check this comment here.

shubhammehra4: @shekhar-rudder adding user_main_id in ids (or any other entity id) is required to add the feature to featute_view

shekhar-rudder: As discussed, we should add it in the constructor
https://rudderlabs.slack.com/archives/C047LGLTP44/p1716794643841299

RuLeKiLLeR: @shekhar-rudder Can you point out where should I make the change?

RuLeKiLLeR: And what that change is

shekhar-rudder: In the constructor `LLMModel`, we need to set `self.ids` so that customers don't need to provide it in the yaml. The value of id will be same as what is currently in the yaml. You can check the installed `profiles_rudderstack` package to get type hints

shekhar-rudder: Had a call with Shubham, entity build spec is currently not available in constructor. So for now we will keep the values in the constructor by making certain assumptions. If those assumptions are not correct for certain customers, they should be able to override the default ids in the code via yaml. Ideally the model should be able to set ids in the execute function

shekhar-rudder (approved):

PR #336, opened 2024-05-28
Changes in the code to fetch details of the custom table

RuLeKiLLeR:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -14,16 +18,36 @@ class LLMModel(BaseModelType):
...
         "additionalProperties": False,
     }
     DEFAULT_LLM_MODEL = "llama2-70b-chat"
 
     def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> None:
+        if "ids" not in build_spec:
+            build_spec["ids"] = []
+
+            build_spec["ids"].append(
+                {"select": "user_main_id", "type": "rudder_id", "entity": "user"}
```
shekhar-rudder, line 49: Here we are assuming that the `entity` will always be user which is not correct. The entity name should come from `entity_key` field of the `model_spec`

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -14,16 +18,36 @@ class LLMModel(BaseModelType):
...
         "required": ["prompt", "prompt_inputs"] + EntityKeyBuildSpecSchema["required"],
         "additionalProperties": False,
     }
     DEFAULT_LLM_MODEL = "llama2-70b-chat"
 
     def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> None:
+        if "ids" not in build_spec:
+            build_spec["ids"] = []
+
+            build_spec["ids"].append(
```
shekhar-rudder, line 48: Add a TODO comment - `select` should be computed from the `entity` object

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
             )
         return super().validate()
 
 
 class LLMModelRecipe(PyNativeRecipe):
     def __init__(self, build_spec: dict) -> None:
         self.prompt = build_spec["prompt"]
-        self.prompt_inputs = build_spec["prompt_inputs"]
+        self.var_inputs = build_spec["var_inputs"] if build_spec["var_inputs"] else None
+        self.table_inputs = (
```
shekhar-rudder, line 105: Since both the fields will never be present in a given model run, the assignment of `var_inputs` and `table_inputs` should be written in a if-else way. It makes it more obvious that only 1 of them will exist

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
         else:
             max_index = max(int(index) for _, index in input_indices)
 
-        if max_index >= len(input_lst):
+        if type(var_inputs) is not type(None) and max_index >= len(var_inputs):
+            raise ValueError(
+                f"Maximum index {max_index} is out of range for var_inputs list."
+            )
+        if type(table_inputs) is not type(None) and max_index >= len(table_inputs):
```
shekhar-rudder, line 94: Duplicating of validation logic in the two if blocks can be avoided. The only difference between the two blocks is the variable (`var_inputs` or `table_inputs`) on which the validation is being applied. So the two blocks can be merged into one, and the variable can be computed with an if/else logic.

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+                    LEFT JOIN predicted_attribute b ON {join_condition}
                 {{% endmacro %}}
                 {{% exec %}} {{{{warehouse.CreateReplaceTableAs(this.Name(), selector_sql())}}}} {{% endexec %}}
             {{% endmacro %}}
             {{% exec %}} {{{{warehouse.BeginEndBlock(begin_block())}}}} {{% endexec %}}"""
+        return query_template
+
+    def register_dependencies(self, this: WhtMaterial):
+        entity_id_column_name = this.model.entity().get("IdColumnName")
+        limit_top_k = (
```
shekhar-rudder, line 159: `limit_top_k` is doing an order by on `frequency` which is defined in `query_template`. Since the two are coupled (change in one will require change in another), this declaration should be moved near to `query_template` declaration

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
-            f"this.DeRef(makePath({self.prompt_inputs[0]}.Model.GetVarTableRef()))"
-        )
-        # Joined_columns used to create a comma seperated string in order to mention
-        # all the columns that are used as input in the query.
-        joined_columns = ", ".join(input_columns)
-        # Join condition to join the predicted column to the original table in order to mention
-        # all the column that are being used in the input columns list.
-        join_condition = " AND ".join([f"a.{col} = b.{col}" for col in input_columns])
-
+    def query_template_360(
```
shekhar-rudder, line 120: `360` can be dropped from the name since it is not being used in 360

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+        self,
+        var_table_ref,
```
shekhar-rudder, line 122: var_table_ref -> input_material_template

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+                    col = "' ||" + table_columns[index] + "|| ' "
+                    prompt_replaced = prompt_replaced.replace(placeholder, col)
+                else:
+                    self.logger.error(
+                        f"Index {index} out of range for input_columns list."
+                    )
+            dereferencing_string = f"{table_name}/var_Table/{entity_id_column_name}"
+            column = this.de_ref(dereferencing_string)
+            table_reference_string = f"{table_name}/var_Table"
+            var_table_ref = f"this.DeRef('{table_reference_string}')"
```
shekhar-rudder, line 236: The names are a bit inconsistent with pb terminology. `<model_name>_model_ref` should be used for the path string

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+                    placeholder = f"{{{word}[{index}]}}"
+                    col = "' ||" + table_columns[index] + "|| ' "
+                    prompt_replaced = prompt_replaced.replace(placeholder, col)
+                else:
+                    self.logger.error(
+                        f"Index {index} out of range for input_columns list."
+                    )
+            dereferencing_string = f"{table_name}/var_Table/{entity_id_column_name}"
+            column = this.de_ref(dereferencing_string)
+            table_reference_string = f"{table_name}/var_Table"
```
shekhar-rudder, line 235: `Table` -> `table`

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+                if 0 <= index < len(table_columns):
+                    placeholder = f"{{{word}[{index}]}}"
+                    col = "' ||" + table_columns[index] + "|| ' "
+                    prompt_replaced = prompt_replaced.replace(placeholder, col)
+                else:
+                    self.logger.error(
+                        f"Index {index} out of range for input_columns list."
+                    )
+            dereferencing_string = f"{table_name}/var_Table/{entity_id_column_name}"
+            column = this.de_ref(dereferencing_string)
```
shekhar-rudder, line 234: column -> entity_id_column_material

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+                index = int(index)
+                if 0 <= index < len(table_columns):
+                    placeholder = f"{{{word}[{index}]}}"
+                    col = "' ||" + table_columns[index] + "|| ' "
+                    prompt_replaced = prompt_replaced.replace(placeholder, col)
+                else:
+                    self.logger.error(
+                        f"Index {index} out of range for input_columns list."
+                    )
+            dereferencing_string = f"{table_name}/var_Table/{entity_id_column_name}"
```
shekhar-rudder, line 233: `dereferencing_string` and `table_reference_string` can be inlined
https://refactoring.com/catalog/inlineVariable.html

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+        elif self.table_inputs:
+            table_column_name = this.model.name()
+            table_columns = []
+            for var in self.table_inputs:
+                column_name = var.get("select")
+                if column_name:
+                    table_columns.append(column_name)
+                else:
+                    print(
+                        "Invalid entry found in prompt_inputs: 'select' key is missing"
```
shekhar-rudder, line 216: This should be part of validate function. By the time the code reaches here, it should be safe to assume that the `select` key is present

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+                    )
+            table_name = self.table_inputs[0].get("from")
```
shekhar-rudder, line 218: Since we are not supporting multiple tables, we should check this in the validate method that multiple tables are not being passed

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+                placeholder = f"{{{word}[{index}]}}"
+                col = "' ||" + input_columns_vars[index] + "|| ' "
+                prompt_replaced = prompt_replaced.replace(placeholder, col)
+
+            # Joined_columns used to create a comma seperated string in order to mention
+            # all the columns that are used as input in the query.
+            joined_columns = ", ".join(input_columns_vars)
+            # Join condition to join the predicted column to the original table in order to mention
+            # all the column that are being used in the input columns list.
+            join_condition = " AND ".join(
```
shekhar-rudder, line 191: `joined_columns` and `join_condition` declaration should happen inside `query_template_360` since they are coupled (similar to `limit_top_k`)

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +87,169 @@ def validate(self):
...
+                raise Exception("no entity found for model")
+            input_columns_vars = [
+                f"{{{{{var}.Model.DbObjectNamePrefix()}}}}"  # Assuming var is already in the format "user.Var('profession')"
+                for var in self.var_inputs
+            ]
+            prompt_replaced = self.prompt.replace(
+                "'", "''", -1
+            )  # This is to escape single quotes
+            input_indices = re.findall(r"{(\w+)\[(\d+)\]}", prompt_replaced)
+            for word, index in input_indices:
```
shekhar-rudder, line 180: Computation of `column_name`, `input_indices` and  `prompt_replaced` is being duplicated in both if-else block. We need to get rid of this duplication

src/predictions/profiles_mlcorelib/py_native/llm.py
RuLeKiLLeR: how to put a this.model.name() in a defined variable name like table_reference_string

shekhar-rudder: How is `this.model.name()` is relevant to this comment ? This comment is about renaming the variables since they are inconsistent with the terminologies

RuLeKiLLeR: To fetch the <model_name>_model_ref model name here i need to do this.model.name() right. Do you mean I should assume it. Can you provide me an example I will put that here.

samples/llm_model/models/profiles.yaml
```diff
@@ -13,17 +13,27 @@ models:
     model_spec:
       prompt: "What is generally the most common gender of the name {input[0]}? Male, Female, or Unknown? Return only a single word, nothing else."
       entity_key: user
-      prompt_inputs:
-        - user.Var('name')
-      llm_model_name: "mistral-7b"
+      var_inputs:
+#        - user.Var('name')
```
shekhar-rudder, line 17: Commented yaml can be removed

samples/llm_model/models/profiles.yaml
```diff
@@ -13,17 +13,27 @@ models:
...
+      table_inputs:
+        - select: first_name
+          from: inputs/rsIdentifies
+      llm_model_name: "llama2-70b-chat"
+      run_for_top_k_distinct:
```
shekhar-rudder, line 22: What is the value here for `run_for_top_k_distinct` ?

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -14,17 +18,39 @@ class LLMModel(BaseModelType):
...
+                        "select": {"type": "string"},
+                        "from": {"type": "string"},
+                    },
+                    "required": ["select", "from"],
+                },
+            },
             "llm_model_name": {"type": ["string", "null"]},
+            "run_for_top_k_distinct": {"type": ["integer", "null"]},
         },
-        "required": ["prompt", "prompt_inputs"] + EntityKeyBuildSpecSchema["required"],
```
shekhar-rudder, line 21: `prompt_inputs` shouldn't be removed. JSON schema supports specifying either of the 2 fields (`var_inputs` and `table_inputs`) being required. We should use that here.

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -14,17 +18,39 @@ class LLMModel(BaseModelType):
...
         "additionalProperties": False,
     }
     DEFAULT_LLM_MODEL = "llama2-70b-chat"
 
     def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> None:
+        if "ids" not in build_spec:
+            build_spec["ids"] = []
+            entity_key = build_spec["entity_key"]
+            # TODO - select should be computed from the entity object
+            build_spec["ids"].append(
```
shekhar-rudder, line 49: The ids can be directly assigned the value instead of doing it in 2 steps (empty array + append)

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -14,17 +18,39 @@ class LLMModel(BaseModelType):
...
+                {"select": "user_main_id", "type": "rudder_id", "entity": entity_key}
```
shekhar-rudder, line 50: `user` is still hardcoded in the `select` field. Its value should be `entity_key + "_main_id"`

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +89,161 @@ def validate(self):
...
             raise ValueError(
-                f"Maximum index {max_index} is out of range for input_columns list."
+                f"Maximum index {max_index} is out of range for the inputs list."
             )
+
+        if table_inputs:
+            for var in table_inputs:
+                input_column_name = var.get("select")
+                if not input_column_name:
+                    raise ValueError("Missing 'select' statement in table_inputs.")
```
shekhar-rudder, line 103: Isn't this case already enforced by `BuildSpecSchema` ?

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +89,161 @@ def validate(self):
...
+        if table_inputs:
+            from_tables = set()
+            for var in table_inputs:
+                table_name = var.get("from")
+                if table_name:
+                    from_tables.add(table_name)
+                else:
+                    raise ValueError("Missing 'from' statement in table_inputs.")
```
shekhar-rudder, line 111: > Isn't this case already enforced by BuildSpecSchema ?

Same here for `from` field

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +89,161 @@ def validate(self):
...
+
+            if len(from_tables) > 1:
+                raise ValueError(
+                    "Multiple different table names found in table_inputs."
```
shekhar-rudder, line 115: `Multiple different table names found in table_inputs.` -> `The model doesn't support specifying more than 1 input tables`

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +89,161 @@ def validate(self):
...
+
         return super().validate()
 
 
 class LLMModelRecipe(PyNativeRecipe):
     def __init__(self, build_spec: dict) -> None:
         self.prompt = build_spec["prompt"]
-        self.prompt_inputs = build_spec["prompt_inputs"]
+        if build_spec["var_inputs"]:
+            print("var_inputs")
```
shekhar-rudder, line 125: `print` can be remvoed

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +89,161 @@ def validate(self):
...
+            predicted_column_name = column_name
+            table_columns = []
+            for var in self.table_inputs:
+                input_column_name = var.get("select")
+                if input_column_name:
+                    table_columns.append(input_column_name)
+
+            table_name = self.table_inputs[0].get("from")
+            entity_id_column_material = this.de_ref(
+                f"{table_name}/var_Table/{entity_id_column_name}"
```
shekhar-rudder, line 234: `T` -> `t`

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +89,161 @@ def validate(self):
...
+                self,
+                var_table_material_template,
+                column_name,
+                entity_id_column_name,
+                input_columns_vars,
+            )
+            self.sql = this.execute_text_template(query_template)
+
+        elif self.table_inputs:
+            predicted_column_name = column_name
```
shekhar-rudder, line 225: No need to create the variable `predicted_column_name`

src/predictions/profiles_mlcorelib/py_native/llm.py
```diff
@@ -62,80 +89,161 @@ def validate(self):
...
+            table_columns = []
+            for var in self.table_inputs:
+                input_column_name = var.get("select")
+                if input_column_name:
```
shekhar-rudder, line 229: if check is not required

samples/llm_model/models/profiles.yaml
RuLeKiLLeR: There are two cases where we can give the value of this parameter or we can leave this empty. So that is why i kept this empty to show that still the model will work.

src/predictions/profiles_mlcorelib/py_native/llm.py
RuLeKiLLeR: Should I remove this then?

src/predictions/profiles_mlcorelib/py_native/llm.py
RuLeKiLLeR: Yes from field has a specific error being raised

samples/llm_model/models/profiles.yaml
shekhar-rudder: We need not test this. No customer will have a field with empty values

src/predictions/profiles_mlcorelib/py_native/llm.py
shekhar-rudder: Can you share the yaml snippet for which you got this error ? I don't think the correct thing was being tested here

src/predictions/profiles_mlcorelib/py_native/llm.py
shekhar-rudder: This error should be handled by `BuildSpecSchema`. No need to do it separately

samples/llm_model/models/profiles.yaml
RuLeKiLLeR: Removing this field shows an error, even though its not in the required field.

shekhar-rudder: What is the error ?

src/predictions/profiles_mlcorelib/py_native/llm.py
shekhar-rudder: The model spec should have either `var_inputs` or `table_inputs` and not both

samples/llm_model/models/profiles.yaml
RuLeKiLLeR: got the error now we don't need it

src/predictions/profiles_mlcorelib/py_native/llm.py
RuLeKiLLeR: this is the error if i don't define a empty ids object
r child folder models: error listing models: error building model llm_gender_prediction: calling pythonClient.ModelFactory: rpc error: code = Unknown desc = Traceback (most recent call last):
  File "/Users/admin/Rudder/lib/python3.10/site-packages/profiles_rudderstack/service.py", line 107, in ModelFactory
    wht_model_ref, py_model_ref = model_type_ref["factory_func"](
  File "/Users/admin/Rudder/lib/python3.10/site-packages/profiles_rudderstack/project.py", line 25, in factory
    model = model_class(
  File "/Users/admin/Rudder/lib/python3.10/site-packages/profiles-sf-cortex/sf_cortex_model.py", line 808, in __init__
    build_spec["ids"].append(
KeyError: 'ids'

shekhar-rudder: The error is because you are appending to a non-existent key.

> The ids can be directly assigned the value instead of doing it in 2 steps

Direct assignment means you use `=` instead of `append`

RuLeKiLLeR: Oh 😅 got the point.

src/predictions/profiles_mlcorelib/py_native/llm.py
RuLeKiLLeR: Made the changes

shekhar-rudder (approved):

PR #356, opened 2024-06-28
 feat: integrate py native logger with predictions code 

shekhar-rudder:
## Description of the change

The logger which is being used in the predictions code doesn't work in the pynative flow since n python subprocess gets created. To make it work instead of getting the default logger at import time, now the code is getting the value at run time using the `.get()` method

- The logger class has a `set_logger` method which is being used by pynative training and prediction classes so that the code can start using the pynative logger
- Since the pynative logger doesn't have a `warning` method and `warn` method in python logger is deprecated. So replaced `warning` with `error` call
- Pynative logger can't be imported in snowflake stored procedure so using print statement there

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/predictions/profiles_mlcorelib/utils/logger.py
+++ b/src/predictions/profiles_mlcorelib/utils/logger.py
@@ -1,28 +1,19 @@
 import logging
 
-class LoggerClass():
 
+class LoggerClass:
     def __init__(self):
         self._logger = self.default_logger()
 
     def default_logger(self):
-        # Define the logger with the desired name
         logger = logging.getLogger("churn_prediction")
-
-        # Configure the logger with the desired log level and handlers
         logger.setLevel(logging.DEBUG)
-
-        # file_handler = logging.FileHandler("classifier.log")
-        # file_handler.setLevel(logging.INFO)
-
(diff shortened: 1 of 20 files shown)


src/predictions/profiles_mlcorelib/utils/logger.py
```diff
@@ -1,21 +1,28 @@
...
-# file_handler.setFormatter(formatter)
-console_handler.setFormatter(formatter)
+    def set_logger(self, newLogger):
+        self._logger = newLogger
 
-# logger.addHandler(file_handler)
-logger.addHandler(console_handler)
+
+# Don't use "warning" method of the logger since it is not impelmented by the pynative logger
+logger = LoggerClass()
```
dpatchigolla, line 28: Can we make this `logger = LoggerClass().get()`? So we don't need so many changes in all other files?

shekhar-rudder: If you call `get()` in this file only, then the updated value from `set_logger` won't be reflected

dpatchigolla (approved):

PR #380, opened 2024-07-22
Changed the sql query to add features

RuLeKiLLeR:
## Description of the change

Changed the sql query to add features

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/predictions/profiles_mlcorelib/py_native/attribution_report.py
+++ b/src/predictions/profiles_mlcorelib/py_native/attribution_report.py
@@ -342,16 +342,11 @@ class AttributionModelRecipe(PyNativeRecipe):
                     if value_flag
                     else ""
                 )
+                + f"""coalesce(coalesce(cost, 0) / nullif(coalesce({conversion_name}_first_touch_count, 0), 0),0) AS {conversion_name}_first_touch_cost_per_conv,
+                                    coalesce(coalesce(cost, 0) / nullif(coalesce({conversion_name}_last_touch_count, 0), 0),0) AS {conversion_name}_last_touch_cost_per_conv,"""
                 + (
-                    f"""coalesce({conversion_name}_first_touch_count, 0) / nullif(coalesce(cost, 0), 0) AS {conversion_name}_first_touch_cost_per_conv,
-                                    coalesce({conversion_name}_last_touch_count, 0) / nullif(coalesce(cost, 0), 0) AS {conversion_name}_last_touch_cost_per_conv,"""
-                    if value_flag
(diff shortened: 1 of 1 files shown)


src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -342,6 +342,19 @@ def _get_final_selector_sql(
                     if value_flag
                     else ""
                 )
+                + (
+                    f"""coalesce({conversion_name}_first_touch_count, 0) / nullif(coalesce(cost, 0), 0) AS {conversion_name}_first_touch_cost_per_conv,
+                                    coalesce({conversion_name}_last_touch_count, 0) / nullif(coalesce(cost, 0), 0) AS {conversion_name}_last_touch_cost_per_conv,"""
+                    if value_flag
```
dpatchigolla, line 348: Why is there a value_flag check for cost_per_conv? Shoudln't it be only for return on ad spend? Looks like the else condition has the same string again? What's the purpose of the if-else block here

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -342,6 +342,19 @@ def _get_final_selector_sql(
...
+                    else f"""coalesce({conversion_name}_first_touch_count, 0) / nullif(coalesce(cost, 0), 0) AS {conversion_name}_first_touch_cost_per_conv,
+                                    coalesce({conversion_name}_last_touch_count, 0) / nullif(coalesce(cost, 0), 0) AS {conversion_name}_last_touch_cost_per_conv,"""
+                )
+                + (
+                    f"""coalesce({conversion_name}_first_touch_conversion_value, 0) / nullif(coalesce(cost, 0), 0) AS {conversion_name}_first_touch_roas,
```
dpatchigolla, line 353: Same as above. This should be cost/value, not value/cost;

Also both should handle scenario of  denominator having 0. If denominator is zero, it should just default to zero.

dpatchigolla (changes requested):

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
RuLeKiLLeR: The idea was to wait for your direction on where to use the value flag that is why added the same string in both if-else. Now you have provided me with more context hence I have formatted it properly now.

dpatchigolla (approved):

PR #386, opened 2024-07-24
changing BuildSpecSchema to support new yaml

joker2411:
## Description of the change

> The required yaml structure is changed. This is to support that. Ticket [Link](https://linear.app/rudderstack/issue/PRML-784/update-model-with-the-new-yaml).

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -26,38 +29,54 @@ class AttributionModel(BaseModelType):
     BuildSpecSchema = {
         "type": "object",
         "properties": {
-            **EntityKeyBuildSpecSchema["properties"],
```
dpatchigolla, line 29: I think this is default `entity_key` for a model right? Why remove this?

dpatchigolla: Shouldn't we have this in place of conversion/properties/Entity_key? Can you check with Shashank if that's feasible to define `EntityKeyBuildSpecSchema` in a nested object? Else any disadvantage in not using this? 
There's a notion of model entity, which lets us define user.Var() in the spec where user is the entity of the model. I think that may be lost if we don't use it.

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -26,38 +29,54 @@ class AttributionModel(BaseModelType):
...
+                            "required": ["from", "timestamp"],
+                        },
+                    },
+                    CONVERSION_VARS: {
+                        "type": "array",
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "name": {"type": "string"},
+                                "timestamp": {"type": "string"},
```
dpatchigolla, line 53: Are `user.Var("first_paid_date")` etc counted as strings?

dpatchigolla (commented):
Can you address the comments? Also can you update the wyze repo too and create a parallel pr with this new structure after you run it locally?

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
joker2411: yes.

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
joker2411: checked with Shashank. we need to define `entity_key` in model_spec directly to use user.Var() syntax. 
Moving the entity_key to model_spec

dpatchigolla (approved):

PR #407, opened 2024-08-07
Added the functionality of conversion_window in conversion_vars

RuLeKiLLeR:
## Description of the change

Added the functionality of conversion_window in conversion_vars

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -107,6 +108,9 @@ class AttributionModel(BaseModelType):
     }
 
     def __init__(self, build_spec: dict, schema_version: int, pb_version: str) -> None:
+        for conversion_var in build_spec[CONVERSION][CONVERSION_VARS]:
+            if "conversion_window" not in conversion_var:
+                conversion_var["conversion_window"] = "90d"
```
dpatchigolla, line 113: _In the build spec schema, add a new key conversion_window for each conversion in conversions_var. This should be optional._ 

_~If it's not provided by the user, it should get a default value of 90d.~_

This requirement got changed in the beginning itself. Default should be No conversion window, in which case, there should be no filter on the datediff between first/last touch date and conversion date

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -501,6 +541,27 @@ def register_dependencies(self, this: WhtMaterial):
         cte_query_list, conversion_name_list, value_flag_list = list(), list(), list()
         for conversion_info in conversion_vars:
             conversion_name = conversion_info["name"]
+            conversion_window = conversion_info["conversion_window"]
+            match = re.match(r"(\d+)([mhd])", conversion_window)
```
dpatchigolla, line 545: Is there a reason you modified the regex from the ticket `r'^\d+[mhd]$'`?That would match the word exactly while this would allow text before and after the intended duration (ex: `30d days` would pass this regex)

image

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -501,6 +541,27 @@ def register_dependencies(self, this: WhtMaterial):
...
+                raise ValueError(
+                    "Invalid conversion window format. Use formats like 30m, 2h, 7d."
+                )
+
+            quantity = int(match.group(1))
+            granularity = match.group(2)
+
+            # Convert conversion window to days for SQL comparison
+            if granularity == "m":
+                conversion_window_days = quantity  # 1 day = 1440 minutes
```
dpatchigolla, line 556: `conversion_window_days` can be outside the if-else conditions as it's the same in all conditions. You can also keep it `n_conversion_window` to make it applicable to hours and minutes too.

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -501,6 +541,27 @@ def register_dependencies(self, this: WhtMaterial):
...
+                conversion_flag = "minute"
+            elif granularity == "h":
+                conversion_window_days = quantity  # 1 day = 24 hours
+                conversion_flag = "hour"
```
dpatchigolla, line 560: Can you rename `conversion_flag` to `conversion_granularity`

dpatchigolla (changes requested):

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
RuLeKiLLeR: So should i remove this here, because then I will have to make changes in the below script as well

RuLeKiLLeR: Removed it

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
RuLeKiLLeR: Changed it back. I was using the different regex to divide the string into parts

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -142,35 +146,82 @@ def _create_with_query_template(
...
+                (
+                    {journey_query}
+                ) AS journey
+                ON conversion_tbl.{entity_id_column_name} = journey.{entity_id_column_name} AND journey.timestamp <= conversion_tbl.converted_date
+            ), 
+            {conversion_name}_user_view AS (
+                SELECT 
+                    {entity_id_column_name},
                     """
+            + (
```
dpatchigolla, line 186: There's too much of duplication in this blocks. 
Can you manage the if-else conditions outside and create those strings, then add them here? 
Something like this (DON'T ADD IT DIRECTLY - TEST IT AND MODIFY IF REQUIRED):

```
def _generate_date_case(field, conversion_flag):
    if conversion_flag:
        return f"""CASE 
            WHEN DATEDIFF({conversion_granularity}, {field}_timestamp, converted_date) <= {conversion_window} THEN DATE({field}_timestamp) 
            ELSE NULL 
        END AS {field}_date"""
    else:
        return f"DATE({field}_timestamp) AS {field}_date"

def _generate_value_case(field, conversion_flag, value_flag):
    if conversion_flag and value_flag:
        return f"""CASE 
            WHEN DATEDIFF({conversion_granularity}, {field}_timestamp, converted_date) <= {conversion_window} THEN {field}_conversion_value 
            ELSE NULL 
        END AS {field}_conversion_value"""
    elif value_flag:
        return f"{field}_conversion_value AS {field}_conversion_value"
    else:
        return ""

def generate_query_part():
    query_parts = []
    
    for attribution_type in ['first_touch', 'last_touch']:
        query_parts.append(_generate_date_case(attribution_type, conversion_flag))
        query_parts.append(f"{attribution_type}_{campaign_id_column_name}")
        query_parts.append(_generate_value_case(attribution_type, conversion_flag, value_flag))
    
    return ",\n".join(filter(None, query_parts))

query_part = generate_query_part()

```

dpatchigolla (changes requested):

PR #408, opened 2024-08-08
Changing yaml structure

joker2411:
## Description of the change

> Ticket [Link](https://linear.app/rudderstack/issue/PRML-778/yaml-cleanup).

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/predictions/profiles_mlcorelib/ml_core/preprocess_and_predict.py
+++ b/src/predictions/profiles_mlcorelib/ml_core/preprocess_and_predict.py
@@ -51,7 +51,7 @@ def preprocess_and_predict(
         results = json.load(f)
     train_model_id = results["model_info"]["model_id"]
     stage_name = results["model_info"]["file_location"]["stage"]
-    model_name = results["model_info"]["file_location"]["file_name"]
+    pkl_model_file_name = results["model_info"]["file_location"]["file_name"]
     model_hash = results["config"]["material_hash"]
     input_model_name = results["config"]["input_model_name"]
 
@@ -145,8 +145,8 @@ def preprocess_and_predict(
         model = trainer.load_model(filename)
         return model
 
-    def predict_helper(df, model_name: str) -> Any:
-        trained_model = load_model(model_name)
+    def predict_helper(df, pkl_model_file_name: str) -> Any:
+        trained_model = load_model(pkl_model_file_name)
         df.columns = [x.upper() for x in df.columns]
         return trainer.predict(trained_model, df)
 
@@ -171,7 +171,8 @@ def preprocess_and_predict(
             ),
             input_types=[PandasDataFrameType(types)],
             input_names=features,
-            imports=[f"{stage_name}/{model_name}.pkl"] + connector.delete_files,
+            imports=[f"{stage_name}/{pkl_model_file_name}.pkl"]
+            + connector.delete_files,
             packages=constants.SNOWFLAKE_TRAINING_PACKAGES + ["cachetools==4.2.2"],
         )
         class predict_scores:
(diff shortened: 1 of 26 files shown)


src/predictions/profiles_mlcorelib/ml_core/preprocess_and_predict.py
```diff
@@ -45,13 +45,13 @@ def preprocess_and_predict(
...
     """
 
-    model_file_name = constants.MODEL_FILE_NAME
     connector.compute_udf_name(model_path)
 
     with open(model_path, "r") as f:
         results = json.load(f)
     train_model_id = results["model_info"]["model_id"]
     stage_name = results["model_info"]["file_location"]["stage"]
+    model_name = results["model_info"]["file_location"]["file_name"]
```
dpatchigolla, line 54: `"file_name": f"{trainer.output_profiles_ml_model}_{model_file_name}",`

This seems to be what's in file_name, and it has suffix `rs_predictions_trained_model`. 
How is it model_name?

tests/integration/utils.py
```diff
@@ -111,19 +118,19 @@ def assert_training_artefacts():
         {
             "regex": "Material_traininG_model_.+",
             "reports": [
-                "01-feature-importance-chart-ltv_classification",
-                "02-test-lift-chart-ltv_classification",
-                "03-test-pr-auc-ltv_classification",
-                "04-test-roc-auc-ltv_classification",
+                "01-feature-importance-chart",
```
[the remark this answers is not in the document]
dpatchigolla: Is this related to current pr, or are you correcting an existing mistake?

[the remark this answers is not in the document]
joker2411: no, we didn't remove the model name suffix in outputs. This change is mainly because of following reasons:

1. previously model_name suffix which was added was a constant thing(almost). Now it's changing in case of pynative depending on the model_name.
2. We were anyways using this output name dictionary in integration test to match the naming `pattern`, not the exact names. So, in case if the model_name is different, prefix of figures should be same.

Therefore now this is adjusted to support both python and pynative model. (original results still have model_name as suffix)

src/predictions/profiles_mlcorelib/ml_core/preprocess_and_predict.py
joker2411: so, it was working in previous structure because it has same model_name for pynative training and prediction as it was coming from `model_config.yaml`

Now, `trainer.output_profiles_ml_model` would hold different values in pynative training model run and pynative predictions model run because current running model is different in both cases.

So, previous structure would have only supported python models but this would support both python and pynative models.

src/predictions/profiles_mlcorelib/ml_core/preprocess_and_train.py
```diff
@@ -52,9 +50,13 @@ def train_and_store_model_results(
             "connector and trainer are required in kwargs for training in train_and_store_model_results"
         )
 
-    model_file = connector.join_file_path(
-        f"{trainer.output_profiles_ml_model}_{model_file_name}"
-    )
+    pkl_model_file_name = kwargs.get("pkl_model_file_name", None)
+    if pkl_model_file_name is None:
```
dpatchigolla, line 54: Can you combine this with the above ValueError? You can check for all Nones at once.

src/predictions/profiles_mlcorelib/ml_core/preprocess_and_train.py
```diff
@@ -52,9 +50,13 @@ def train_and_store_model_results(
             "connector and trainer are required in kwargs for training in train_and_store_model_results"
         )
 
-    model_file = connector.join_file_path(
-        f"{trainer.output_profiles_ml_model}_{model_file_name}"
-    )
+    pkl_model_file_name = kwargs.get("pkl_model_file_name", None)
```
dpatchigolla, line 53: You don't need None as 2nd param. That's the default for `.get()`

dpatchigolla (approved):
Approved with one minor suggestion.

PR #412, opened 2024-08-13
Removed additional dependencies and imports and included new imports

RuLeKiLLeR:
…in requirements.txt

## Description of the change

Removed additional dependencies and imports and included new imports

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


conda_env.yml
```diff
@@ -21,6 +21,9 @@ dependencies:
     - scipy==1.7.1
     - seaborn==0.11.2
     - matplotlib
+    # Google
+    - google-auth
```
dpatchigolla, line 25: cc: @shekhar-rudder

requirements.txt
```diff
@@ -28,4 +28,10 @@ setuptools>=68.2.2
 # pyOpenSSL>=24.0.0
 # cryptography>=42.0.4
 pillow>=10.2.0
-pycaret==3.3.0
\ No newline at end of file
+pycaret==3.3.0
+
+# Additional imports
+google-auth
```
dpatchigolla, line 34: @shekhar-rudder can you check if these are required? I guess python model will fail on UI right after merge if these are not installed there.

src/predictions/profiles_mlcorelib/config/model_configs.yaml
```diff
@@ -66,7 +66,7 @@ train:
           - ada     # AdaBoost Classifier
           - gbc     # Gradient Boosting Classifier
         regressors :
-          - xgboost # Xgboost Regressor
+          - xgboost # Xgboost Classifier
```
dpatchigolla, line 69: This is regressor

tests/integration/requirements.txt
```diff
@@ -25,7 +25,12 @@ db-dtypes==1.2.0
 pycaret==3.3.1
 
 # not directly required, pinned by Snyk to avoid a vulnerability
-setuptools==68.2.2 
+setuptools==68.2.2
 pyOpenSSL==24.0.0
 cryptography==42.0.4
 pillow==10.2.0
+
+google-auth
```
dpatchigolla, line 33: These are not required.

dpatchigolla (commented):
To get approval from @shekhar-rudder reg packages in requirements.txt and setup.py

conda_env.yml
shekhar-rudder: Should we get rid of this file ?

requirements.txt
shekhar-rudder: - I don't see `google-auth` and `google-auth-oauthlib` being used anywhere in the code. Why are we adding them ?
- `boto3` is an indirect dependency of `pandas-redshift` and `redshift-connector`. So till now we didn't see any error even without an explicit dependency. But we should definitely add it. The current UI version is `1.34.153`. We can keep the constraint same as [redshift-connector](https://github.com/aws/amazon-redshift-python-driver/blob/64cbd54ef6a5e71d98ce193630d09267cc154379/requirements.txt#L4)
- There is no need of adding `kubernetes` package since that code flow is not being used.

RuLeKiLLeR: In this fine you will find the import for google-auth-oauthlib import : src/predictions/profiles_mlcorelib/connectors/wh/bigquery_base_connector.py

conda_env.yml
dpatchigolla: @RuLeKiLLeR you can delete this file.

requirements.txt
RuLeKiLLeR: yes
though we might not need google-auth for that but google-auth-oauthlib is necessary for that import

shekhar-rudder: Ok then lets get rid of `google-auth`. And pin a version range for `google-auth-oauthlib`

shekhar-rudder: `google-auth-oauthlib>=1.0.0,<2.0.0` The latest version of the package is 1.2.1. Why are we capping it with 2.0.0 ?

RuLeKiLLeR: Oh ok I kept it so that if a later version comes even that will be downloaded. Let me change it then.

requirements.txt
```diff
@@ -28,4 +28,8 @@ setuptools>=68.2.2
 # pyOpenSSL>=24.0.0
 # cryptography>=42.0.4
 pillow>=10.2.0
-pycaret==3.3.0
\ No newline at end of file
+pycaret==3.3.0
+
+# Additional imports
+google-auth-oauthlib>=1.0.0
+boto3>=0.0.1
```
shekhar-rudder, line 35: https://github.com/rudderlabs/rudderstack-profiles-classifier/pull/412#discussion_r1721445970

shekhar-rudder (approved):

PR #419, opened 2024-08-14
Made all the changes need to support path/ref format for label columns

RuLeKiLLeR:
## Description of the change

Made all the changes need to support path/ref format for label columns

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/samples/py_native/models/profiles.yaml
+++ b/samples/py_native/models/profiles.yaml
@@ -13,7 +13,7 @@ models:
         - packages/feature_table/entity/user/n_orders_completed
       ml_config:
         data:
-          label_column: is_churned_7_days
+          label_column: packages/feature_table/entity/user/is_churned_7_days
           label_value: 1
           prediction_horizon_days: 7
           eligible_users: '*'
@@ -33,7 +33,7 @@ models:
       ml_config:
         data:
           task: regression
-          label_column: days_since_last_seen
+          label_column: packages/feature_table/entity/user/days_since_last_seen
           prediction_horizon_days: 7
           eligible_users: '*'
           output_profiles_ml_model: ltv_regression_integration_test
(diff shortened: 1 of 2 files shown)


src/predictions/profiles_mlcorelib/py_native/prediction.py
```diff
@@ -103,7 +103,10 @@ def describe(self, this: WhtMaterial):
     def register_dependencies(self, this: WhtMaterial):
         this.de_ref(self.build_spec["training_model"])
         for input in self.build_spec["inputs"]:
-            this.de_ref(input)
+            if input == self.build_spec["ml_config"]["data"]["label_column"]:
```
dpatchigolla, line 106: This doesn't check for label_column if input doesn't have that. 

You don't need if-else here - it's same as doing following:
```
if a == b:
   list.append(b)
else:
   list.append(a)
```
Always doing list.append(a)  will be the same.

dpatchigolla (changes requested):

src/predictions/profiles_mlcorelib/py_native/prediction.py
shekhar-rudder: > This doesn't check for label_column if input doesn't have that.

Doing deRef separately for inputs and label_column even if there are common elements is okay. We need not have a conditional

dpatchigolla: @RuLeKiLLeR keep your for loop simple, no if conditions  and then outside of the for loop, you can just add `this.de_ref(self.build_spec["ml_config"]["data"]["label_column"])`.

dpatchigolla (approved):

PR #430, opened 2024-08-24
Added a new has_cost flag to check for the cost in campaign_details

RuLeKiLLeR:
## Description of the change

Added a new has_cost flag to check for the cost in campaign_details

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -397,6 +417,27 @@ def _get_final_selector_sql(
         from_query = f"""
                         FROM index_cte a"""
         for conversion_name, value_flag in zip(conversion_name_list, value_flag_list):
+            # select_query = (
```
dpatchigolla, line 420: Can you remove the commented code?

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -350,6 +350,26 @@ def _get_index_cte(
                                 """
         return index_cte_query
 
+    def _query_cost_calculated_fields(
+        self, conversion_name: str, value_flag: bool
```
dpatchigolla, line 354: Where are you using the self.has_cost variable here?

dpatchigolla (changes requested):

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
RuLeKiLLeR: I had not completed my code. Now you may check again
You can check the final output here:
select * from AMBUJ_TEMP.Material_campaign_performance_report_22ffdffa_180;

dpatchigolla: can you make it 

```
if self.has_cost:
     # actual logic
     return cost_fields
return ""

```

This is more pythonic

src/predictions/profiles_mlcorelib/py_native/attribution_report.py
```diff
@@ -590,6 +606,17 @@ def register_dependencies(self, this: WhtMaterial):
         user_journeys = self.config[CONVERSION][TOUCHPOINTS]
         conversion_vars = self.config[CONVERSION][CONVERSION_VARS]
         campaign_details = self.config[CAMPAIGN][CAMPAIGN_DETAILS]
+        self.has_cost = any(
+            "cost" in detail
```
dpatchigolla, line 610: `"cost" in detail`
Why are you checking for substring match? It should be exact match only. This will return True for `cost_combined` for example, but it will fail later because there's no `cost` key when query in `_query_cost_calculated_fields` gets executed

dpatchigolla (changes requested):

dpatchigolla (approved):

PR #472, opened 2024-10-08
using dataclass for inputs

joker2411:
## Description of the change

> Ticket [Link](https://linear.app/rudderstack/issue/PRML-845/use-data-classespydantic-models-for-material-and-model-objects).

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/profiles_mlcorelib/ml_core/preprocess_and_predict.py
```diff
@@ -346,11 +346,20 @@ def predict_scores_rs(df: pd.DataFrame) -> pd.DataFrame:
 
     model_path = os.path.join(output_dir, args.json_output_filename)
 
+    inputs_info: List[utils.InputsConfig] = []
+    try:
+        if isinstance(args.inputs, list) and isinstance(args.inputs[0], dict):
```
dpatchigolla, line 351: What would be the else condition? If we expect it should always go to the if condition, let's skip the condition? And let it error out in case of unexpected behavior? Currently it will silently make mistakes if the instance checks fail and that's not intended.

src/predictions/profiles_mlcorelib/wht/pythonWHT.py
```diff
@@ -522,11 +522,11 @@ def get_complete_sequences(sequences: List[Sequence]) -> int:
             feature_data_min_date_diff,
         )
 
-    def get_latest_seq_no(self, inputs: List[dict]) -> int:
-        return int(inputs[0]["table_name"].split("_")[-1])
+    def get_latest_seq_no(self, inputs: List[utils.InputsConfig]) -> int:
+        return int(inputs[0].table_name.split("_")[-1])
 
     def get_inputs(self, selector_sqls: str, skip_compile: bool) -> List[dict]:
```
dpatchigolla, line 528: The output datatype here would be List[InputsConfig] right? This is the first place where we are creating the inputconfig, so lets just add this here. Other places it's okay if we miss occasionally.

dpatchigolla (approved):

PR #475, opened 2024-10-09
Feature/prml 972 optimise  get material names function

dpatchigolla:
## Description of the change

These optimisations brought down the time taken for fetching past materials from 15 mins to 3 mins in a test project

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 


src/predictions/profiles_mlcorelib/wht/pythonWHT.py
```diff
@@ -255,16 +265,21 @@ def _get_material_names(
...
             inputs[0]["model_name"],
             inputs[0]["model_hash"],
             start_time,
             end_time,
             prediction_horizon_days,
+            registry_table,
         )
 
         materials = list()
+        material_validity_cache = {}
```
shekhar-rudder, line 282: Why not make `cache` a member variable of the connector class instead of passing it around as an argument ?

dpatchigolla: Connector class needs to be pickled and re-created inside a different sub-process, within snowflake stored procedure or in local for Redshift/BigQuery. 

This caused pickling errors when I tried to do the same with the registry table. So kept the cache scoped within a method instead of making it an object level parameter.

shekhar-rudder: > This caused pickling errors when I tried to do the same with the registry table.

The pickling errors will be only with the registry table, rt ? The cache which is a simple dict can be pickled.

dpatchigolla: Fair point, let me test it.

src/predictions/profiles_mlcorelib/wht/pythonWHT.py
```diff
@@ -106,6 +106,8 @@ def _validate_historical_materials_hash(
         self,
         input: dict,
         material_seq_no: Optional[int],
+        material_registry_table=None,
+        material_validity_cache: dict = None,
```
shekhar-rudder, line 110: Cache argument can be removed from this file

shekhar-rudder (approved):

PR #484, opened 2024-10-15
Disabled logging for pycaret setup and added n_jobs = 1 so it is hono…

dpatchigolla:
…red in parallel execution of tune_model

## Description of the change

Currently pycaret tries to write to log file in root directory and it fails when it doesnt get write access. fixed the issue by disabling logging, and making n_jobs = 1 so tune_model doesn't spawn more processes without this context (a limitation in pycaret)

A customer reported this issue on their snowflake. I couldn't replicate the failure in our snowflake, but could see that on our snowflake instance, it does try to write to a log file on root directory, and that gets gracefully handled. I could replicate the failure locally though, making the `logs.log` file read only. Even if i disable logging in setup() method, it still tries to log to the same file in tune_models in the parallel processing step (the setup config is not getting honored by tune_model). Only solution I could find was to disable parallel processing, by giving n_jobs = 1 in setup. this doesn't create new parallel jobs with default configs. 

> Description here

## Type of change
- [x] Bug fix (non-breaking change that fixes an issue)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/predictions/profiles_mlcorelib/connectors/RedshiftConnector.py
+++ b/src/predictions/profiles_mlcorelib/connectors/RedshiftConnector.py
@@ -51,10 +51,18 @@ class RedshiftConnector(CommonWarehouseConnector):
         session.execute(f"SET search_path TO {self.schema};")
         return session
 
+    def _run_query(self, query: str) -> redshift_connector.cursor.Cursor:
+        try:
+            return self.session.execute(query)
+        except:
+            # Retry by recreating session to handle timeouts
+            self.session = self.build_session(self.creds)
+            return self.session.execute(query)
+
     def run_query(self, query: str, response=True) -> Optional[List]:
         """Runs the given query on the redshift connection and returns a Named Tuple."""
         if response:
-            query_run_obj = self.session.execute(query)
+            query_run_obj = self._run_query(query)
             if query_run_obj.description:
                 column_names = [desc[0] for desc in query_run_obj.description]
                 row_outputs = query_run_obj.fetchall()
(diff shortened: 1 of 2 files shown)


src/predictions/profiles_mlcorelib/trainers/MLTrainer.py
```diff
@@ -237,6 +237,10 @@ def _train_model(
             categorical_features=categorical_cols,
             fold=n_folds,
             fold_strategy=fold_strategy,
+            system_log=False,
```
shekhar-rudder, line 240: Can we add some of the PR description as code comments ?

shekhar-rudder (approved):

src/predictions/profiles_mlcorelib/trainers/MLTrainer.py
dpatchigolla: Done.

src/predictions/profiles_mlcorelib/connectors/RedshiftConnector.py
```diff
@@ -51,10 +51,18 @@ def build_session(self, credentials: dict) -> redshift_connector.cursor.Cursor:
         session.execute(f"SET search_path TO {self.schema};")
         return session
 
+    def _run_query(self, query: str) -> redshift_connector.cursor.Cursor:
```
shekhar-rudder, line 54: ` Addressed Redshift timeout issue that came up because we removed parallel processing in pycaret tune_model` -> Will removing parallel processing cause the runs to be slower?

dpatchigolla: Yes, the training part.

dpatchigolla: I added that in the comment too - in the prev commit. We can address the issue when someone reports a problem with it.

PR #569, opened 2025-01-27
Picking eligible users automatically

joker2411:
## Description of the change

> Ticket [Link](https://linear.app/rudderstack/issue/PRML-1061/pick-eligible-users-automatically).

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

### DIFF
--- a/src/predictions/profiles_mlcorelib/connectors/Connector.py
+++ b/src/predictions/profiles_mlcorelib/connectors/Connector.py
@@ -315,6 +315,10 @@ class Connector(ABC):
         pass
 
     @abstractmethod
+    def get_eligible_users(self, feature_table, input_column_types, **kwargs):
+        pass
+
+    @abstractmethod
     def call_procedure(self, *args, **kwargs):
         pass
 
--- a/src/predictions/profiles_mlcorelib/connectors/SnowflakeConnector.py
+++ b/src/predictions/profiles_mlcorelib/connectors/SnowflakeConnector.py
@@ -21,6 +21,7 @@ from snowflake.snowpark.session import Session
 from ..utils import utils
 from ..utils import constants
 from ..utils.logger import logger
+from ..trainers.MLTrainer import MLTrainer
 from ..connectors.Connector import Connector
 from ..wht.rudderPB import MATERIAL_PREFIX
 from cryptography.hazmat.primitives.serialization import load_pem_private_key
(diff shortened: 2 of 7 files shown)


src/predictions/profiles_mlcorelib/ml_core/preprocess_and_train.py
```diff
@@ -96,17 +98,9 @@ def prepare_feature_table(
...
-                feature_table_name, filter_condition=trainer.eligible_users
-            )
-        else:
-            default_user_shortlisting = (
-                f"{trainer.label_column} != {trainer.label_value}"
-            )
-            feature_table = connector.get_table(
-                feature_table_name, filter_condition=default_user_shortlisting
-            )
+        feature_table = connector.get_table(
```
dpatchigolla, line 101: By this point, eligible_users always exist right? Can you add a wrapper function on top of get_table and call it `get_filtered_table` where filter_condition is mandatory? 
The condition can be `*` too, but can't be None.

src/predictions/profiles_mlcorelib/py_native/propensity.py
```diff
@@ -167,6 +167,10 @@ def _get_prediction_spec(self, training_model_ref: str, training_spec) -> dict:
                 )
         if self.build_spec["prediction"].get("eligible_users", None) is not None:
```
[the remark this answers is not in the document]
dpatchigolla: also for both train and predict, can you make a check that an empty string in yaml is stored as None? 
You are relying on `config.get("eligible_users") is None` condition in the preprocess_and_train.py; But `"" is None` would return false. So if customer gives a blank string, that should be treated as None explicitly, and lets do that at the entry point itself - in this code.

src/predictions/profiles_mlcorelib/ml_core/preprocess_and_train.py
```diff
@@ -135,13 +129,22 @@ def preprocess_and_train(
...
-    trainer = kwargs.get("trainer", None)
+    connector: Connector = kwargs.get("connector", None)
+    trainer: MLTrainer = kwargs.get("trainer", None)
     min_sample_for_training = constants.MIN_SAMPLES_FOR_TRAINING
     cardinal_feature_threshold = constants.CARDINAL_FEATURE_THRESHOLD
     train_config = merged_config["train"]
-
     feature_table = None
+
+    logger.get().info("Getting eligible users condition:")
```
dpatchigolla, line 139: This line isn't required, when the eligible_users is not None - if customer gives something. insert a different log inside the if condition.

src/predictions/profiles_mlcorelib/ml_core/preprocess_and_train.py
```diff
@@ -135,13 +129,22 @@ def preprocess_and_train(
...
+    if trainer.eligible_users is None:
+        trainer.eligible_users = connector.get_default_eligible_users_condition(
```
dpatchigolla, line 141: You can add a logger inside here
"eligible_users flag not given. Trying to auto-select a subset of users on which to train the model"

src/predictions/profiles_mlcorelib/connectors/Connector.py
```diff
@@ -314,6 +314,146 @@ def join_file_path(self, file_name: str) -> str:
     def run_query(self, query: str, **kwargs):
         pass
 
+    def get_default_eligible_users_condition(
```
dpatchigolla, line 317: This should be in trainer, and not in connector. Specially in ClassificationTrainer. 
we should also have one in RegressionTrainer. Lets create that as a follow-up ticket and discuss how to implement that. Can you create the ticket?

src/predictions/profiles_mlcorelib/connectors/Connector.py
```diff
@@ -314,6 +314,146 @@ def join_file_path(self, file_name: str) -> str:
...
+            logger.get().error(
+                "Label_column is required to be binary to fetch default eligible users for training."
+            )
+            raise Exception(
+                "Label_column is required to be binary to fetch default eligible users for training."
+            )
+
+        # For each boolean feature, trying both True and False values
+        for bool_feature in booleantype_features:
+            for bool_value in [True, False]:
```
dpatchigolla, line 394: Does this logic work when values of a column is 1/0, y/n etc? And in all warehouses?

dpatchigolla: This looks a bit fragile. We are making multiple assumptions here - 
1. Features should be recognised as boolean type
2. The values should be True/False, not 1/0, y/n etc
3. We are not testing on the label column in the feature table (current default will never be picked up in this)
4. There can be nulls in the binary features. The null can be in eligible users flag too, or not - but ensure the label column split is correctly handled. ex - `eligible_users: is_payer = 0 or is_payer is null` can be a potential eligible users condition. Also, when you are running queries/converting to dataframes and checking the result, if nulls aren't explicitly checked for, the label distribution ignores the nulls completely. Make sure you consider this (ex: select  is_churned, count(*) from c360 where is_payer != 1 may ignore is_payer= null condition - beware of that)

Can you rewrite this part to ensrue all these are addressed

src/predictions/profiles_mlcorelib/connectors/Connector.py
dpatchigolla: When you do this, please add a few unit-tests to ensure this works well. prepare a few sample datasets and ensure the eligible_users are picked correctly.

dpatchigolla (changes requested):

src/predictions/profiles_mlcorelib/connectors/RedshiftConnector.py
```diff
@@ -81,7 +81,7 @@ def run_query(self, query: str, **kwargs) -> Optional[Union[List, pd.DataFrame]]
                 return query_output
             else:
                 raise Exception(
-                    "No result set is present for given query. Please check the query."
+                    f"No result set is present for given query. Please check the query: {query}."
```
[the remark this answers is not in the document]
joker2411: this was already being handled in case of BigQuery. Therefore added the changes for redshift and snowflake.

src/predictions/profiles_mlcorelib/connectors/SnowflakeConnector.py
```diff
@@ -103,10 +103,13 @@ def run_query(
...
-        if return_type == "dataframe":
-            return query_run_obj
-        return query_run_obj.collect()
+        try:
+            query_run_obj = self.session.sql(query)
+            if return_type == "dataframe":
+                return query_run_obj
+            return query_run_obj.collect()
+        except Exception as e:
+            raise Exception(f"Couldn't run the query: {query}. Error: {str(e)}")
```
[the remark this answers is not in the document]
joker2411: While writing the sql queries and testing them, faced this issue. Therefore added proper logs

src/predictions/profiles_mlcorelib/connectors/Connector.py
joker2411: created.

src/predictions/profiles_mlcorelib/connectors/Connector.py
```diff
@@ -40,6 +41,17 @@ def validate_sql_table(self, table_name: str, entity_column: str) -> None:
                 f"SQL model table {table_name} has duplicate values in entity column {entity_column}. Please make sure that the column {entity_column} in all SQL model has unique values only."
             )
 
+    def get_filtered_table(self, feature_table_name, filter_condition):
+        if filter_condition is None:
+            raise Exception(
```
dpatchigolla, line 46: Why is the Exception twice? This seems to be an error. 
Can you write unit-tests for the PR? For all the new changes, lets make sure we have the tests - especially as this is a slightly complex PR.

dpatchigolla: For this specific test, we should add tests for a case where it would not have the filter_condition and we expect the exception to raise properly (This wouldn't probably cause an issue though because of two raise exceptions).

dpatchigolla (commented):
This still seems to be in progress? Main changes around moving the core logic to trainer and changing the logic itself seems to be still in progres. I added comments only till changes so far.

